Issue #5, opened 2024-10-13
🐛 [BUG] - Not viewed counter doesn't display large number 


majestyck:
### Description

When there is a big number, it doesn't get displayed correctly. Here for example, it should display 390 but hasn't enough space To do it

This issue doesn't occur on horizontal layout

### Screenshots

![Screenshot_2024-10-13-11-31-48-01_8932e2da1be34319ec0a30b28399191a](https://github.com/user-attachments/assets/8d2c19d3-6611-4d67-9eb3-f45435ee58ca)

### Platform

Android

### App Version

0.2.3

PR #8, opened 2024-10-13
[Bugfix] Fix display large numbers

DenixJG:
Fix the error when display large numbers #5 

Example: 
image

The numbers in the example image was random generated.


PartyDonut (approved):
Looks good to me, thanks for the help!

And congratulations on being the first contributor.

Issue #6, opened 2024-10-13
🐛 [BUG] - Status bar get merged with the content


majestyck:
### Description

When you scroll down into a page, there is no padding to prevent the status bar to be merged with the content 

### Reproduction steps

```bash
1. Go to any page
2. Scroll down a little
3. The status bar will be merged into the content
```

### Screenshots

![Screenshot_2024-10-13-11-25-37-60_8932e2da1be34319ec0a30b28399191a](https://github.com/user-attachments/assets/c4fab4d1-3ba3-4f8f-9023-37a70ad7ecdc)
![Screenshot_2024-10-13-11-26-35-67_8932e2da1be34319ec0a30b28399191a](https://github.com/user-attachments/assets/1968a0d8-5b0c-4c78-af51-dab09fa20db5)

### Platform

Android

### App Version

0.2.3

PR #9, opened 2024-10-13
[Bugfix] Wrap body in SafeArea for improved layout

DenixJG:
Wrapped the `body` widget in a `SafeArea` widget to prevent content from being obscured by notches, status bars, or other screen intrusions. #6

### DIFF
--- a/lib/screens/shared/detail_scaffold.dart
+++ b/lib/screens/shared/detail_scaffold.dart
@@ -85,149 +85,151 @@ class _DetailScaffoldState extends ConsumerState<DetailScaffold> {
         },
         backgroundColor: Theme.of(context).colorScheme.surface,
         extendBodyBehindAppBar: true,
-        body: Stack(
-          children: [
-            SingleChildScrollView(
-              physics: const AlwaysScrollableScrollPhysics(),
-              child: Stack(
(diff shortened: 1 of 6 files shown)


PartyDonut:
Thanks for the PR. 
While this would fix the issue, some of the overlapping of the statusbar is by design. For desktop (and probably mobile in some cases) it would look strange.
Wrapping the nestedScaffold with a safe area changes padding in pretty much the entire app. Preferable this would only fix the problem for the sync_item_details.dart pop-up.

screenshot

PartyDonut (changes requested):
Thanks for the PR. 
While this would fix the issue, some of the overlapping of the statusbar is by design. For desktop (and probably mobile in some cases) it would look strange.
Wrapping the nestedScaffold with a safe area changes padding in pretty much the entire app. Preferable this PR would only fix the problem for the sync_item_details.dart pop-up.

screenshot

DenixJG:
Thank you for the feedback!

I had considered that the overlapping might have been intentional, but I do agree that it can feel a bit off in some sections. I'll take a closer look at adjusting the fix specifically for the sync_item_details.dart pop-up to avoid affecting other areas of the app.

PR #16, opened 2024-10-14
Add libmpv-dev to Linux workflow to fix compilation

bazzadazza72:
Hi there,

I managed to boil down the compilation issues to only Android and iOS, the former returning the error `Cannot use @TaskAction annotation on method IncrementalTask.taskAction$gradle_core() because interface org.gradle.api.tasks.incremental.IncrementalTaskInputs is not a valid parameter to an action method.` (the Kotlin version was also out of date, spewing dependency errors into the run log, that has since been fixed, I hope), and the latter returning the `No Team Found in Archive` error during the IPA export. The main build workflow has been deleted and replaced with separate workflows for compiling on desktop, mobile and GitHub Pages.

Feel free to test out my changes and if you have any issues, let me know. If you still want to collate the GH actions into a single file, I'd be happy to do it once the Android and iOS compilation issues are fixed.

<sub>Sorry for the amount of commits and the long rant. I was doing some experimenting, and there were some things that worked and some that didn't. This is my first foray into app development and there were some spots where I didn't know what I was doing.</sub>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -198,38 +198,38 @@ jobs:
           name: ${{ env.PACKAGE_NAME }}
           path: build/macos/Build/Products/Release/${{ env.PACKAGE_NAME }}.dmg
 
-  # build-linux:
-  #     runs-on: ubuntu-latest
+  build-linux:
+      runs-on: ubuntu-latest
 
-  #     steps:
-  #       - name: Checkout repository
-  #         uses: actions/checkout@v4.1.1
+      steps:
+        - name: Checkout repository
+          uses: actions/checkout@v4.1.1
 
-  #       - name: Set up Flutter
-  #         uses: subosito/flutter-action@v2.16.0
-  #         with:
-  #           channel: ${{ vars.FLUTTER_CHANNEL }}
-  #           flutter-version: ${{ vars.FLUTTER_VERSION }}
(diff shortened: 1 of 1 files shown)


PartyDonut (changes requested):
Thanks for taking your time to fix the build issues. 

What issues did you encounter with the Android build? 
For this repository it works fine at the moment.

Lets also include just the changes for the linux part of the build.yaml. Considering that the build will (or should atleast) always upload all the platforms.

bazzadazza72:
The only issue I encountered during the Android build was the error I mentioned in my first comment, other than the out-of-date Kotlin version.

I was performing my builds in a fork I created, without any secrets or environment variables, so that very well could be my issue.

I'll fix the Linux deployment, and if there's anything else you want me to change, I'm all ears.

bazzadazza72:
Sorry 'bout that, I definitely did not mean to close that.

Anywho, I've added `libmpv-dev` as a dependency to the Linux build, again, if there's anything else you need done, I'm all ears.

PartyDonut (approved):

PartyDonut:
Thanks for the help, for things to pick-up you can always try picking up one of the issues or features (not sure how experienced you are?)

Otherwise testing out the app and reporting bugs always helps.

Issue #53, opened 2024-10-19
🐛 Missing Shared Library on Fedora 40


Nanianmichaels:
### Description

Hello.

I have just run Fladder on Fedora 40, and it errored out with a missing shared library, more specifically libmpv.so.1.

Please consider adding that library to the app's lib folder, or adding instructions on how to get it on the system on the download page (included below for F40).

### Reproduction steps

1. Try to load the app via terminal.
2. App will error out from missing shared library; if loaded by double-clicking the executable, it will simply not load.

For F40 users, you need to install at least mpv-devel, and then symlink the shared library.

Steps required:

- sudo dnf install mpv-devel
- sudo ln -s /usr/lib64/libmpv.so /usr/lib64/libmpv.so.1

I personally also installed mpv, but the library in question was only available on mpv-devel.

For Debian-based systems, the commands apparently are:
sudo apt install libmpv-dev libmpv2 (or libmpv1, depending on the source)
sudo ln -s /usr/lib/x86_64-linux-gnu/libmpv.so /usr/lib/libmpv.so.1

I have not, however, tested the solution on a Debian-based system.

### Platform

Linux

### App Version

v0.2.6-alpha

bazzadazza72:
> I have not, however, tested the solution on a Debian-based system.

I believe the issue happens on all Linux distributions, including Debian. I'll have a look at what modules it uses, could be a typo somewhere.

bazzadazza72:
If I can't fix the issue, or another workaround other than the one you've provided is found, I'll write some documentation

PartyDonut:
I think it would also help if we can package the linux binaries in something like Flatpak or an appimage (or both)?

That would include all libraries I believe.

bazzadazza72:
I did think about doing that, I would recommend doing an AppImage, personally, but if you want to distribute it to a wider audience, Flatpak would be the way to go.

PartyDonut:
Let's try for both if possible 👍🏼 .

bazzadazza72:
Just found this as a resource: https://appimage-builder.readthedocs.io/en/latest/examples/flutter.html, very promising

bazzadazza72:
> Let's try for both if possible

Will get started on that this week, shouldn't take too long 🤞

lmgarret:
I'm working on a Flatpak build [here](https://github.com/lmgarret/Fladder/tree/lmg/flatpak) , I'm getting close to something clean for a PR. 

[...]

Issue #122, opened 2024-11-05
💡 Upgrade media player


PartyDonut:
### Description

Currently the app uses media_kit.
This is fine for most scenarios and easy to use but it has some small drawbacks for certain platforms.

Option 1:
Upgrade to latest development media_kit and possible fine tune the options per platform or even per installment?

Option 2:
Investigate into using https://pub.dev/packages/fvp instead. This might yield better result.

Could maybe even be a mix of the two.

### Reproduction steps

df

### Screenshots

![DESCRIPTION](LINK.png)

### Platform

iOS

bazzadazza72:
Using `fvp` might be a good plan. Reading the documentation leads me to believe it uses FFmpeg instead of MPV as its backend, which might mitigate the library issues we've been having on Linux.

PartyDonut:
fvp uses libmdk, with a fall-back to ffmpeg if there is no GPU decoding available. But yes, especially the Dolby Vision support has me hopefull.
One of the big reasons not to switch is I have no idea how subtitle support is for fvp.

PR #125, opened 2024-11-06
Automatically build a Flatpak bundle

lmgarret:
## Pull Request Description

This adds the manifest file to build a flatpak bundle with each release. It does not cover publishing to Flathub, which is a different topic IMO.

Heavily inspired by the [Jellyfin Media Player flatpak manifest](https://github.com/flathub/com.github.iwalton3.jellyfin-media-player/blob/master/com.github.iwalton3.jellyfin-media-player.json), building ffmpeg, mpv and other dependencies by source as recommended for flatpaks.

## Issue Being Fixed

It is a potential solution to #53

## Checklist

- [X Check that any changes are related to the issue at hand.

### DIFF
--- a/flatpak/nl.jknaapen.fladder.yaml
+++ b/flatpak/nl.jknaapen.fladder.yaml
@@ -1,0 +1,216 @@
+id: nl.jknaapen.fladder
+runtime: org.gnome.Platform
+runtime-version: '46'
+sdk: org.gnome.Sdk
+
+command: Fladder
+finish-args:
+  # X11/Wayland access
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  # Audio playback
+  - --socket=pulseaudio
+  # Network access for Jellyfin streaming
+  - --share=network
+  # File access for downloads/media
+  - --filesystem=home
+  # Hardware acceleration
+  - --device=dri
+  # Allow notification access
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.PowerManagement
+  - --talk-name=org.freedesktop.ScreenSaver
+  - --talk-name=org.freedesktop.portal.*
+
+modules:
+  - name: mpv
+    buildsystem: meson
+    config-opts:
+      - -Dlibmpv=true
+      - -Dcplayer=false
+      - -Dbuild-date=false
+      - -Dmanpage-build=disabled
+      - -Dvaapi=enabled
+      - -Dcuda-hwaccel=enabled
+      - -Dpulse=enabled
+      - -Dalsa=enabled
+      - -Duchardet=enabled
+    cleanup:
+      - /lib/pkgconfig
+      - /share
+      - /include
+    sources:
+      - type: git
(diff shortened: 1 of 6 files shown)


linux/my_application.cc
```diff
@@ -109,6 +109,7 @@ static void my_application_init(MyApplication *self) {}
 
 MyApplication *my_application_new()
 {
+	g_set_prgname(APPLICATION_ID);
```
lmgarret, line 112: This fixes KDE showing the default wayland icon in some scenarios (e.g. alt+tab) instead of the app's one.
Took me way too much time to find out, ([thanks to this comment](https://github.com/flutter/flutter/issues/53229#issuecomment-2347348554)).

flatpak/nl.jknaapen.fladder.yaml
```diff
@@ -0,0 +1,216 @@
...
+sdk: org.gnome.Sdk
+
+command: Fladder
+finish-args:
+  # X11/Wayland access
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  # Audio playback
+  - --socket=pulseaudio
```
PartyDonut, line 13: Would this mean it only works on pulseaudio, not with pipewire?

PartyDonut (approved):
Looks good to me. had a small question but feel free to merge.
Thanks for adding flatpak support 👍🏼

flatpak/nl.jknaapen.fladder.yaml
lmgarret: That's a fair point, apparently it works if there's the pipewire-pulseaudio compat layer, but we may need to do something like done in this PR https://github.com/flathub/com.github.iwalton3.jellyfin-mpv-shim/pull/10 .
I'll investigate it later

lmgarret: I added the changes to allow mpv's access to the pipewire socket. I can confirm with `pw-cli ls Node` that the previous flatpak would use `client.api=pipewire-pulse`, whereas with this change it uses pipewire directly. Both cases worked as expected.

I've also added a title to the media_kit `PlayerConfiguration` so that the app shows as `Fladder` in the volume mixer and not the default `mpv`. The icon is still mpv's in the volume mixer, I could not find what causes it. But I guess it may not be worth digging into, especially with #122

PartyDonut: Thanks for implementing/checking this out. Always a good idea to support multiple use cases.

> I've also added a title to the media_kit `PlayerConfiguration` so that the app shows as `Fladder` in the volume mixer and not the default `mpv`. The icon is still mpv's in the volume mixer, I could not find what causes it. But I guess it may not be worth digging into, especially with #122

Correct best to wait after that issue is closed whether this is something we want/need fixed.

PartyDonut (dismissed):

PartyDonut (approved):

PartyDonut:
@lmgarret 
This is done/ready to merge I think?

lmgarret:
@PartyDonut looks done to me too yes

PR #126, opened 2024-11-06
bugfix: Fix sending NowPlayingQueue in progress reports

### LINKED ISSUE #121: 🐛 Progress report causes to thread pool starvation on server side
### Description

Hi! This is a bug I've only found now, I've mostly been using the app to play content offline.

While playing some TV show episode, the Jellyfin server became unresponsive. I could see on the server that the count of jellyfin ffmpeg threads went from ~20-30 to serveral hundreds.

I wrote this quick script to monitor the number of ffmpeg threads on the server, polling it every 5s

```bash
#!/bin/bash

while true; do
	current_time=$(date +"%Y-%m-%d %H:%M:%S")
	count=$(pgrep -wf ffmpeg | wc -l)
	echo "$current_time: $count"
	sleep 5
done
```

The jellyfin server logs show:

```log
[2024-11-04 09:46:10.079 +00:00] [WRN] As of "11/04/2024 09:46:08 +00:00", the heartbeat has been running for "00:00:01.2073446" which is longer than "00:00:01". This could be caused by thread pool starvation.
```

### Reproduction steps

1. On the server sidem, run the shared script above to monitor the number of ffmpeg threads started
2. Play an episode of a TV show with a lot of episodes (around 170 in my case)
3. Monitor the number of threads, which should quickly rise, around 10 per sec in my case. The server may eventually crash

See `Other notes` below for more investigation details which brought me to these reproduction steps

### Screenshots

DESCRIPTION

## Other notes
I initially got the issue while running the Linux app. To investigate the API calls being made, I ran the web version and got the same issue. I tried streaming the same episode using the Jellyfin web interface and the official Jellyfin Media Player and did not run in this issue.

Inspecting the API calls, I noticed that:

1. Fladder calls `POST /Sessions/Playing/Progress` every 1.5s when the offical frontend do it every ~10s
2. Fladder sends a pretty big (170+ in that case) `NowPlayingQueue` list when reporting progress, whereas the official frontend does not send any?

Point 1. sure seems convenient, having a better granular progress tracking (although resuming an episode to max 10s before you stopped watching it can actually be a good feature). But point 2. a.k.a. the big `NowPlayingQueue` looks suspicious. I then tried to

2. 1. Play an episode in a TV show with way less episodes in total. The `NowPlayingQueue` had around ~50 elements, and I did not run into the same issue.
2. 2. Patched the code heartbeat
https://github.com/DonutWare/Fladder/blob/043e465058ae8c65df522f2c98da089ea5956b91/lib/models/playback/transcode_playback_model.dart#L65-L66
and here
https://github.com/DonutWare/Fladder/blob/043e465058ae8c65df522f2c98da089ea5956b91/lib/models/playback/direct_playback_model.dart#L65-L66

to return `[]`. It doesn't trigger the bug as well (although not a fix)

### Platform

Linux, Web

### App Version

v0.4.1

lmgarret:
## Pull Request Description

Fixes a potential thread pool starvation error on the server side when reporting the player's progress, by not sending any `NowPlayingQueue`. The `Next Up` banner/screen still shows up at the end of the content, and I haven't found any side effects.

## Issue Being Fixed

Resolves #121

## Checklist

### DIFF
--- a/lib/models/playback/direct_playback_model.dart
+++ b/lib/models/playback/direct_playback_model.dart
@@ -121,7 +121,6 @@ class DirectPlaybackModel implements PlaybackModel {
             isMuted: false,
             isPaused: false,
             repeatMode: RepeatMode.repeatall,
-            nowPlayingQueue: itemsInQueue,
           ),
         );
     return null;
@@ -138,7 +137,6 @@ class DirectPlaybackModel implements PlaybackModel {
             playSessionId: playbackInfo.playSessionId,
             positionTicks: position.toRuntimeTicks,
             failed: false,
-            nowPlayingQueue: itemsInQueue,
           ),
           totalDuration: totalDuration,
         );
@@ -170,7 +168,6 @@ class DirectPlaybackModel implements PlaybackModel {
         isMuted: false,
         positionTicks: position.toRuntimeTicks,
         repeatMode: RepeatMode.repeatall,
-        nowPlayingQueue: itemsInQueue,
       ),
     );
 
--- a/lib/models/playback/transcode_playback_model.dart
+++ b/lib/models/playback/transcode_playback_model.dart
@@ -122,7 +122,6 @@ class TranscodePlaybackModel implements PlaybackModel {
             isMuted: false,
             isPaused: false,
             repeatMode: RepeatMode.repeatall,
-            nowPlayingQueue: itemsInQueue,
           ),
         );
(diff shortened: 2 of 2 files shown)


PartyDonut (approved):
Looks good, thanks.

Only thing this might affect syncplay (whenever/if) thats ever added. But we will see when we get there.

PR #222, opened 2025-02-09
feature: Create Windows installer

bazzadazza72:
## Pull Request Description

I created a Windows installer using Inno Setup. It includes the GPL license that needs to be accepted before installation. This PR also contains the changes required to build installers when new Fladder versions release.

## Issue Being Fixed

*No issue has been fixed*

## Screenshots / Recordings

*No screenshots required*

## Checklist


bazzadazza72:
Let me know if there's anything else I need to change before merging.

.github/workflows/build.yml
```diff
@@ -101,6 +101,8 @@ jobs:
 
   build-windows:
     runs-on: windows-latest
+    env:
+      FLADDER_VERSION: ${{needs.fetch-info.outputs.version_name}}
```
PartyDonut: This is not needed I believe, its already imported by the needs: [fetch-info] parameter.

windows/windows_setup.iss
```diff
@@ -0,0 +1,124 @@
+[Setup]
+AppId={{D573EDD5-117A-47AD-88AC-62C8EBD11DC7}
+AppName="Fladder"
+AppVersion={%FLADDER_VERSION|latest}
```
PartyDonut, line 4: The current installers title is "Fladder version latest" I'm assuming that is because it can't find this variable?
Might be best to just always use the current version from the fetch-info step if possible.

PartyDonut (changes requested):
Thanks for implementing this so quickly🚀.

Installer works great only some small questions.

1. Can we give the installer some more branding, a simple banner would make sure users can see they are installing Fladder instead of some generic application.
2. The uninstaller icon Installer-Visuals now shows as the default program can we change this to the Fladder icon as wel.
3. When uninstalling is there a option to ask if the user wants to remove or keep any generated user data? Also if the user has synchronized any videos we can't delete those should we notify the user about this?

bazzadazza72:
> 1. Can we give the installer some more branding, a simple banner would make sure users can see they are installing Fladder instead of some generic application.

That sounds like a good idea, I think it looks a bit generic too.

> 2. The uninstaller icon Installer-Visuals now shows as the default program can we change this to the Fladder icon as wel.

Yep, I'll do that. I was experimenting with it, because according to their documentation, it looks like it can be provided a few icon sizes. With the current icon, I think it'll display with the main part being too small.

> 3. When uninstalling is there a option to ask if the user wants to remove or keep any generated user data? Also if the user has synchronized any videos we can't delete those should we notify the user about this?

I'm not sure about this, as the uninstaller usually deletes the files that are placed by the installer, but I can have a look to see if Inno can delete other directories as part of the uninstall.

bazzadazza72:
*Phew*, finally done. You're free to test using [this artifact](https://github.com/bazzadazza72/Fladder/actions/runs/13235942020), lmk if there's anything you need me to add before we merge.

I figured out the app version: I tried to add it the beginning of the job (which didn't work), then I added it to the step directly which DID work 🥳

I didn't mean to reference the issue, I added "#2" to the commit message without knowing that.

windows/windows_setup.iss
```diff
@@ -0,0 +1,131 @@
...
+
+[Icons]
+Name: "{autoprograms}\Fladder"; Filename: "{app}\fladder.exe"
+Name: "{autodesktop}\Fladder"; Filename: "{app}\fladder.exe"; Tasks: desktopicon
+
+[Run]
+Filename: "{app}\fladder.exe"; Description: "{cm:LaunchProgram,Fladder}"; Flags: nowait postinstall skipifsilent
+
+[UninstallDelete]
+Type: filesandordirs; Name: {localappdata}\DonutWare
```
PartyDonut: Looks like this might be the incorrect folder? At least for me it's not deleting anything.

windows/windows_setup.iss
```diff
@@ -0,0 +1,131 @@
...
+
+[Messages]
+ConfirmUninstall=Uninstalling %1 will delete all user data, including your synced items. Are you sure you want to continue?
```
PartyDonut: If possible, could we have the option to toggle whether the user wants to keep the data. Instead of cancelling the installer as a whole?

Also the synced items can be anywhere on the drive. So the user has to delete those themselves. I think deleting those is out of the scope for the installer.

PartyDonut (commented):
Thanks for updates and adding the images looks great 😄. 

I tested the installer and left some comments.
If deletion of userdata is a bit difficult to implement it's fine to leave it out for now 👍🏼 .

windows/windows_setup.iss
bazzadazza72: It should be correct. The directory should resolve to `C:\Users\<username>\AppData\Local\DonutWare`.

bazzadazza72:
I've just added a confirmation where the user has to decide whether to delete that folder and if they decline, the uninstall continues without deletion. I haven't done a CI run yet, lemme do that and I'll edit this comment.

windows/windows_setup.iss
PartyDonut: Aah yeah that resolves to the "cache" directory for windows that's where the crash logs are stored.
The actual application data is stored in `C:\Users\<username>\AppData\Roaming\DonutWare`.

windows/windows_setup.iss
```diff
@@ -0,0 +1,144 @@
...
+[Run]
+Filename: "{app}\fladder.exe"; Description: "{cm:LaunchProgram,Fladder}"; Flags: nowait postinstall skipifsilent
+
+[Code]
+procedure CurUninstallStepChanged(CurUninstallStep: TUninstallStep);
+begin
+  case CurUninstallStep of
+    usUninstall:
+      begin
+        if MsgBox('Would you like to delete your synced items? This action cannot be undone.', mbConfirmation, MB_YESNO) = IDYES then
```
PartyDonut: Could we rename this to.

`Would you like to delete the application's data? This action cannot be undone. Synced files will remain unaffected.`

bazzadazza72: Done. I also added a dialog that tells the user the folder couldn't be deleted if it fails to delete.

PartyDonut (approved):
LGTM, thanks for adding Windows installer support.

PR #314, opened 2025-04-17
Video BufferSize settings for MPV

Julien9969:
## Pull Request Description

Add MPV Player Option to Choose Buffer Size (MB)

The default buffer size for MPV player is set to 32 MB. However, depending on the source file, 32 MB might correspond to less than one minute of video, especially if the playback speed is higher than 1x. This can lead to playback stuttering if the server takes time to send data or if the connection is poor.

image

I've checked for MDK, but the buffer seems to be time-based instead of size-based. This might make it less important for this context.

Also : small fix in [getInitials](lib/util/string_extensions.dart), double space in the actor's full name causes a RangeError when accessing the string index.

## Checklist

- [ X] If a new package was added, did you ensure it works for all supported platforms? Is the package well maintained
- [X ] Check that any changes are related to the issue at hand.


PartyDonut (dismissed):
Hi, thanks for taking your time to help out it's greatly appreciated 😃 .

Looks good to me,  I've deleted the discovery_provider.mapper.dart as it was probably included because of the generation.

PartyDonut (approved):

PR #315, opened 2025-04-19
Fix "Ends at" don't use playrate and next episode change selected stream

Julien9969:
## Pull Request Description

- The "Ends at" on player controls display end time with date + runtime, so it does not consider the playrate. 

- I'v updated the playback speed slider to use a step size of 5% and set the maximum speed at 3, i my opinion 10 was a bit hight and only a little partion of the slider was usefull

- When switching to the next video using player button it reset the streams selection to the default ones, the app now preserves the current audio and subtitle stream indices. This avoids falling back to defaults streams index that return the server.
For sysnced items i can't test it because in `createOfflinePlaybackModel` of **playback_model.dart**, 
```dart
    final children = ref.read(syncChildrenProvider(syncedItem));
    final syncedItems = children.where((element) => element.videoFile.existsSync()).toList();
    final itemQueue = syncedItems.map((e) => e.createItemModel(ref));

```
`ref.read(syncChildrenProvider(syncedItem));` return and empty list even if next episode is download

Also i've added a DEVELOPMENT.md file to help set up the project, including installation steps, Linux dependencies, and usage of a fake server for testing, maybe i'm not doing it the best way but might be usefull.

As the change were few lines i didn't create separate PR, but if you prefer i can do that.

## Checklist

### DIFF
--- a/DEVELOPEMENT.md
+++ b/DEVELOPEMENT.md
@@ -1,0 +1,64 @@
+# 🚀 Fladder Dev Setup
+
+## 🔧 Requirements
+
+Ensure the following tools are installed:
+
+- [Flutter SDK](https://flutter.dev/docs/get-started/install) (latest stable)
+- [Android Studio](https://developer.android.com/studio) (for Android development and emulators)
+- [VS Code](https://code.visualstudio.com/) with:
+  - Flutter extension
+  - Dart extension
+
+Verify your Flutter setup with:
+
+```bash
+flutter doctor
+```
+
+## 🚀 Quick Start
+
+```bash
+# Clone the repository
+git clone https://github.com/DonutWare/Fladder.git
+cd Fladder
+
+# Install dependencies
+flutter pub get
+```
+
+## 🐧 Linux Dependencies
+
+If you're on **Linux**, install the `mpv` dependency:
+
+```bash
+sudo apt install libmpv-dev
+```
+
+## 🛠️ Running the App
+
+1. **Connect a device** or launch an emulator.
+2. In VS Code:
+   - Select the target device (bottom right corner).
+   - Press `F5` or go to **Run > Start Debugging**.
+   - If prompted, select **"Run Anyway"**.
+
+## ⚙️ Code Generation
+
+Generate build files (e.g., for `json_serializable`, `freezed`, etc.):
+
+```bash
+flutter pub run build_runner build
+```
+
+> Tip: Use `watch` for continuous builds during development:
+```bash
+flutter pub run build_runner watch
+```
+
+## 🌐 Using the Fake Server
+To connect to the fake server :
+Add a new server with the following details:
+   - **Server URL**: `http://22b469df.fladder.nl`  
+   - **Username**: `User 1`  
+   - **Password**: `Txnw6RWYb8yEtD`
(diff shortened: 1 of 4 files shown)


DEVELOPEMENT.md
```diff
@@ -0,0 +1,64 @@
...
+
+> Tip: Use `watch` for continuous builds during development:
+```bash
+flutter pub run build_runner watch
+```
+
+## 🌐 Using the Fake Server
+To connect to the fake server :
+Add a new server with the following details:
+   - **Server URL**: `http://22b469df.fladder.nl`
```
PartyDonut, line 62: Development document looks fine, This however I'm not sure., the fake server response was mostly implemented to please the google reviewers. 

Would probably be better to link to the jellyfin demo server?
https://demo.jellyfin.org/stable/web/

lib/models/playback/playback_model.dart
PartyDonut, line 1: I would prefer to leave subtitle handling to the server. 
And if we do implement something like this, to maybe check whether a subtitle with the same language code and type can be selected? Instead of relying on the index.

This would also only work when the files have the exact same amount of subtitles (or at least the same index).

lib/screens/video_player/components/video_player_options_sheet.dart
```diff
@@ -502,9 +502,9 @@ Future<void> showPlaybackSpeed(BuildContext context) {
                           width: 250,
                           child: FladderSlider(
                             min: 0.25,
-                            max: 10,
+                            max: 3,
```
PartyDonut, line 505: This looks fine to me.
However maybe instead of "forcing" the user to a min/max we could have a  -/+ 0.05% buttons that way the user can do whatever they want? Not sure how usefull that would be I personally don't use the functionality alot so not sure if it fits the use case.

PartyDonut (changes requested):
Thanks again for contributing 😄.

Left a few comments/ideas. 

Also could you split up the changes? 
Not so much because of the size, more so because the release notes depends on pull-requests. 
Splitting the PR's up even if they are smaller makes the release notes cleaner and more organized.

Something like this:
1. chore: Added Development.md
2. fix: Ends at calculation when playbackrate is not 1x
3. feat: Videplayer playbackrate changes
4. feat: Videoplayer Audio/Subtitle stream remember

lib/screens/video_player/components/video_player_options_sheet.dart
Julien9969: I think the slider is great. Clicking multiple times on a button is not really user-friendly, and for 99.99% of cases, I think people won't go above 2.

lib/models/playback/playback_model.dart
Julien9969: That's right, it could be an issue I hadn't considered. Does the server know which current stream is selected and send that information? I find it quite annoying that the stream selection is reset when moving to the next video, but I will try to handle it more intelligently

Julien9969:
> Thanks again for contributing 😄.
> 
> Left a few comments/ideas.
> 
> Also could you split up the changes? Not so much because of the size, more so because the release notes depends on pull-requests. Splitting the PR's up even if they are smaller makes the release notes cleaner and more organized.
> 
> Something like this:
> 
> 1. chore: Added Development.md
> 2. fix: Ends at calculation when playbackrate is not 1x
> 3. feat: Videplayer playbackrate changes
> 4. feat: Videoplayer Audio/Subtitle stream remember

Sure, I will split it. I'll temporarily convert this to a draft to keep track of the conversation

lib/models/playback/playback_model.dart
PartyDonut: Lets convert this to a discussion for now, I think there are a lot of things at work that could make this difficult to implement in a way that most people would want.

PR #325, opened 2025-05-04
fix: Ends at calculation when playbackrate is not 1x

Julien9969:
## Pull Request Description

The "Ends at" on player controls display end time with date + runtime, so it does not consider the playrate.

## Checklist


PartyDonut (approved):
LGTM, Thanks for splitting it up 👍🏼

PR #326, opened 2025-05-04
feat: playbackrate changes - max set to 3 and smaller steps

Julien9969:
## Pull Request Description

Updated of the playback speed slider to use a step size of 5% and set the maximum speed at 3, i my opinion 10 was a bit hight and only a little partion of the slider was usefull

## Checklist


PartyDonut (approved):
Thanks!

PR #339, opened 2025-05-15
Feat player track remember when not default

Julien9969:
## Pull Request Description

* When switching to the next video using the player button, the stream selections are reset to the default ones (those provided by the server). The app now tries to preserve the current audio and subtitle selections. This avoids falling back to the default stream indexes returned by the server.
To determine if a track matches the previous one, the app uses the scoring logic implemented in the Jellyfin web client (see: [https://github.com/DonutWare/Fladder/discussions/327#discussioncomment-13100519](https://github.com/DonutWare/Fladder/discussions/327#discussioncomment-13100519)).

* Two options have been added in the player settings to choose whether the current track selections should kept if possible. These settings are linked to the Jellyfin user configuration returned by the server (`UserGet...`), and are stored in the account so the config change will be get on main page refresh.

image

For synced items, I can't test it because in `createOfflinePlaybackModel` from **playback_model.dart**:

```dart
final children = ref.read(syncChildrenProvider(syncedItem));
final syncedItems = children.where((element) => element.videoFile.existsSync()).toList();
final itemQueue = syncedItems.map((e) => e.createItemModel(ref));
```

`ref.read(syncChildrenProvider(syncedItem));` returns an empty list, even if the next episode is downloaded, so it’s currently impossible to switch to the next video from the player.

## Checklist

### DIFF
--- a/lib/models/items/media_streams_model.dart
+++ b/lib/models/items/media_streams_model.dart
@@ -175,6 +175,20 @@ class StreamModel {
   });
 }
 
+class AudioAndSubStreamModel extends StreamModel {
+  final String language;
+  final String displayTitle;
+  AudioAndSubStreamModel({
+    required this.displayTitle,
+    required super.name,
+    required super.codec,
+    required super.isDefault,
+    required super.isExternal,
+    required super.index,
+    required this.language,
+  });
+}
+
 class VersionStreamModel {
   final String name;
   final int index;
@@ -250,19 +264,17 @@ extension SortByExternalExtension<T extends StreamModel> on Iterable<T> {
   }
 }
 
-class AudioStreamModel extends StreamModel {
-  final String displayTitle;
-  final String language;
+class AudioStreamModel extends AudioAndSubStreamModel {
   final String channelLayout;
 
   AudioStreamModel({
-    required this.displayTitle,
+    required super.displayTitle,
     required super.name,
     required super.codec,
     required super.isDefault,
     required super.isExternal,
     required super.index,
-    required this.language,
+    required super.language,
     required this.channelLayout,
   });
 
@@ -292,8 +304,8 @@ class AudioStreamModel extends StreamModel {
 
   AudioStreamModel.no({
     super.name = 'Off',
-    this.displayTitle = 'Off',
-    this.language = '',
+    super.displayTitle = 'Off',
+    super.language = '',
     super.codec = '',
     this.channelLayout = '',
     super.isDefault = false,
@@ -302,19 +314,17 @@ class AudioStreamModel extends StreamModel {
   });
 }
 
-class SubStreamModel extends StreamModel {
+class SubStreamModel extends AudioAndSubStreamModel {
   String id;
   String title;
-  String displayTitle;
-  String language;
   String? url;
   bool supportsExternalStream;
   SubStreamModel({
     required super.name,
     required this.id,
     required this.title,
-    required this.displayTitle,
-    required this.language,
+    required super.displayTitle,
+    required super.language,
     this.url,
     required super.codec,
     required super.isDefault,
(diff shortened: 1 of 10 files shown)


lib/models/items/media_streams_model.dart
```diff
@@ -175,6 +175,20 @@ class StreamModel {
...
+    required this.displayTitle,
+    required super.name,
+    required super.codec,
+    required super.isDefault,
+    required super.isExternal,
+    required super.index,
+    required this.language,
+  });
+}
+
```
Julien9969, line 191: I've create this interface for AudioStreamModel and subStreamModel to avoid duplication of the track selection function. StreamModel doesn't define **language** and **displayTitle** that are needed for score computation (`selectStream`)

lib/providers/service_provider.dart
```diff
@@ -902,4 +902,24 @@ class JellyService {
...
+
+  Future<UserConfiguration?> updateRememberSubtitleSelections() async {
+    final newUserConfiguration = account?.userConfiguration?.copyWith(rememberSubtitleSelections: !(account?.userConfiguration?.rememberSubtitleSelections ?? false));
+
+    var response = await api.usersConfigurationPost(userId: account?.id, body: newUserConfiguration);
+    if (response.isSuccessful && response.statusCode == 200) {
+      return Future.value(newUserConfiguration);
+    }
+    return null;
+  }
```
Julien9969: those doesn't update userconfiguration (204)
i figure out that fladder (swagger) and jellyfin web doesn't use the same post method
image
image

But it seems more to be an issue with the value not having changed.

lib/screens/settings/player_settings_page.dart
```diff
@@ -166,6 +167,28 @@ class _PlayerSettingsPageState extends ConsumerState<PlayerSettingsPage> {
...
+          SettingsListTile(
+            label: Text(context.localized.rememberSubtitleSelections),
+            subLabel: Text(context.localized.rememberSubtitleSelectionsDesc),
+            trailing: Switch(
+              value: ref.watch(userProvider.select(
+                (value) => value?.userConfiguration?.rememberSubtitleSelections ?? true,
+              )),
+              onChanged: (_) => ref.read(userProvider.notifier).setRememberSubtileSelections(),
+            ),
+          ),
```
Julien9969, line 192: I need to add the ontap()

lib/providers/service_provider.dart
Julien9969: Update: 204 is correct, is just mean that the server update the config but response body is empty

lib/models/playback/playback_model.dart
```diff
@@ -201,8 +201,16 @@ class PlaybackModelHelper {
         itemId: firstItemToPlay.id,
         body: PlaybackInfoDto(
           startTimeTicks: startPosition?.toRuntimeTicks,
-          audioStreamIndex: streamModel?.defaultAudioStreamIndex,
-          subtitleStreamIndex: streamModel?.defaultSubStreamIndex,
+          audioStreamIndex: selectAudioStream(
```
PartyDonut: This logic seems to be called twice (both selectAudio and selectSubstream), better to call both once.

lib/models/playback/playback_model.dart
```diff
@@ -411,4 +435,57 @@ class PlaybackModelHelper {
       ref.read(videoPlayerProvider.notifier).loadPlaybackItem(newModel, startPosition: currentPosition);
     }
   }
+
+  int? selectAudioStream(AudioAndSubStreamModel? previousStream, List<AudioAndSubStreamModel>? currentStream, int? defaultStream) {
```
PartyDonut: Let's move these to either a extension in util, or just include it in the AudioAndSubStreamModel class.
So we can keep this file clean (as clean as possible).

PartyDonut (changes requested):
Nice thanks for implementing this.
Code so far looks good to me, works as expected. Left some small change requests, after that I will merge it.

Do we still need to check if the video being played is in the same show/season? Or would the score for different media types (episodes/movies) always be lower then 3?

lib/models/playback/playback_model.dart
Julien9969: streams_selection.dart created, i needed to modify the function as the ref is not avaible outside a the class

PartyDonut (approved):
LGTM, Thanks for implementing this 👍🏼

PR #349, opened 2025-05-20
chore: Changed case naming for Mac OS App

vikingnope:
## Pull Request Description

This PR changes the first letter of the Mac app to be capital instead, this should be the same as the rest of the apps on other OSes

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -214,6 +214,14 @@ jobs:
       - name: Build macOS app
         run: flutter build macos --flavor production --build-number=${{ github.run_number }}
 
+      - name: Ensure correct app name casing
+        run: |
+          APP_DIR="build/macos/Build/Products/Release-production"
+          # Check if the app exists with potentially wrong casing
+          if [ -d "$APP_DIR/fladder.app" ] && [ ! -d "$APP_DIR/Fladder.app" ]; then
+            mv "$APP_DIR/fladder.app" "$APP_DIR/Fladder.app"
+          fi
+
       - name: Create DMG file
         run: hdiutil create -format UDZO -srcfolder build/macos/Build/Products/Release-production/Fladder.app build/macos/Build/Products/Release-production/macOS.dmg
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -215,7 +215,7 @@ jobs:
         run: flutter build macos --flavor production --build-number=${{ github.run_number }}
 
       - name: Create DMG file
-        run: hdiutil create -format UDZO -srcfolder build/macos/Build/Products/Release-production/fladder.app build/macos/Build/Products/Release-production/macOS.dmg
+        run: hdiutil create -format UDZO -srcfolder build/macos/Build/Products/Release-production/Fladder.app build/macos/Build/Products/Release-production/macOS.dmg
```
PartyDonut, line 223: I believe this needs to be renamed to Fladder.app first before adding it to the dmg. The command will work but given that macOS is not case-sensitive it would just move the fladder.app file inside of the dmg archive.

PartyDonut (changes requested):

.github/workflows/build.yml
```diff
@@ -214,8 +214,16 @@ jobs:
       - name: Build macOS app
         run: flutter build macos --flavor production --build-number=${{ github.run_number }}
 
+      - name: Ensure correct app name casing
+        run: |
+          APP_DIR="build/macos/Build/Products/Release-production"
+          # Check if the app exists with potentially wrong casing
+          if [ -d "$APP_DIR/fladder.app" ] && [ ! -d "$APP_DIR/Fladder.app" ]; then
```
PartyDonut: Given that the file is always in lowercase "fladder.app". Just the mv "$APP_DIR/fladder.app" "$APP_DIR/Fladder.app"command is fine no need to check if its lowercase.

Would reduce the complexity.

schembriaiden: No problem, let me change it

PartyDonut (approved):
LGTM, Thanks for helping out 🎉

PR #369, opened 2025-06-05
chore: Added Applications folder for Mac .dmg

vikingnope:
## Pull Request Description
Added a symlink to the Applications folder in the Mac Installer

## Issue Being Fixed

This streamlines the installation process on Mac

## Screenshots / Recordings

Screenshot will be added soon

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -222,8 +222,22 @@ jobs:
           APP_DIR="build/macos/Build/Products/Release-production"
           mv "$APP_DIR/fladder.app" "$APP_DIR/Fladder.app"
 
-      - name: Create DMG file
-        run: hdiutil create -format UDZO -srcfolder build/macos/Build/Products/Release-production/Fladder.app build/macos/Build/Products/Release-production/macOS.dmg
+      - name: Create DMG with drag and drop to Applications
+        run: |
+          # Create temporary directory for DMG contents
+          mkdir -p dmg_temp
+          
+          # Copy the app to the temp directory
+          cp -R "build/macos/Build/Products/Release-production/Fladder.app" dmg_temp/
+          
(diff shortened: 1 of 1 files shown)


PartyDonut:
Thanks for implementing this looks like a good addition. We can merge it as is, however I added a background we could use for the dmg "assets\macos-dmg\Fladder-DMG-Background.jpg". 

Let me know if you are willing to implement this into the build steps using something like [create-dmg](https://github.com/create-dmg/create-dmg).

PartyDonut (dismissed):

schembriaiden:
> Thanks for implementing this looks like a good addition. We can merge it as is, however I added a background we could use for the dmg "assets\macos-dmg\Fladder-DMG-Background.jpg".
> 
> Let me know if you are willing to implement this into the build steps using something like [create-dmg](https://github.com/create-dmg/create-dmg?rgh-link-date=2025-06-07T09%3A15%3A31Z).

Oh that's perfect, I was going to do this but I didn't have a background, I will try implementing it again before you merge this.

schembriaiden:
Could we maybe change the background slightly by adding some kind of different colour at the text parts of the folder and the app so that they are visible, because currently they are not that visible and there is no way to remove the text nor change their colour:

Screenshot 2025-06-07 at 12 52 17

Screenshot 2025-06-07 at 12 53 52

These are some examples:

Screenshot 2025-06-07 at 12 43 56

Screenshot 2025-06-07 at 12 44 32

PartyDonut:
I've changed the background to be lighter, also decreased the drag-target size to not clip into the "applications" text. Let me know if the new version works.

Changes are pushed to develop

schembriaiden:
Okay it should be finished now, let me know what you think 😄

PR #373, opened 2025-06-07
chore: Added nightly workflow

vikingnope:
## Pull Request Description

I thought adding a nightly workflow would benefit this repo so that recent changes made to develop can be tested before being pushed to master/main to be released. This clears up any artifacts after 7 days to create more storage for other workflows.

## Issue Being Fixed

No real issue, just a quality of life improvement.

## Testing

I couldn't test this locally yet using act and docker because some of the builds keep failing, even on the normal `build.yml`, so probably some issue from my end., but I think this new workflow should be fine since it is mostly based on the `build.yml`.

## Checklist


PartyDonut (changes requested):
Like the idea, was thinking about this as well. 
Few things though

1. I think it would be better to build it nightly (at midnight) on develop instead of on every push and closed PR. While not really a problem at the moment I can imagine the action/build points will run out quite fast this way.
2. Creating a separate file would also require us to keep the two files in sync in case anything changes. Might be better to include the nightly build inside of the existing build.yml? 
3.  When this is pushed it might cause the update checker to show a new update is available, despite it being just a nightly release. That's something that needs fixing first probably.

schembriaiden:
These are the changes I have done after your review:

1. I unified the nightly and build workflows since there was a bunch of duplicate code (Should have done that before lol)
2. The nightly workflow runs at midnight everyday but only if there have been any changes to the develop branch in the last 24 hours.
3. Addressed the issue regarding the update checker potentially showing the nightly releases as potential updates

lib/util/update_checker.dart
PartyDonut, line 1: Let's revert this file for now, there are a few issues that would probably need some more testing after a few nightly builds.  Thanks for giving it a shot though 😃

.github/workflows/build.yml
```diff
@@ -6,27 +6,104 @@ on:
       - "v*"
     branches:
       - master
+      - develop
```
PartyDonut: We don't need this for nightly builds right?

.github/workflows/build.yml
```diff
@@ -6,27 +6,104 @@ on:
...
   pull_request:
     paths:
       - pubspec.yaml
       - .github/workflows/build.yml
       - .fvmrc
+    branches:
+      - develop
```
PartyDonut: This one can also be removed

.github/workflows/build.yml
```diff
@@ -6,27 +6,104 @@ on:
...
     types:
       - opened
       - reopened
+      - closed
```
PartyDonut: We don't need to run the build on closed

.github/workflows/build.yml
```diff
@@ -6,27 +6,104 @@ on:
...
+
+      - name: Check if should run and determine build type
+        id: check
+        run: |
+          # Determine build type based on trigger and input
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" && "${{ github.event.inputs.build_type }}" != "auto" ]]; then
+            BUILD_TYPE="${{ github.event.inputs.build_type }}"
+          elif [[ "${{ startsWith(github.ref, 'refs/tags/v') }}" == "true" ]]; then
+            BUILD_TYPE="release"
+          elif [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/develop" ]]; then
```
PartyDonut: Let's remove the push on develop or on merged PR's for simplicity sake better to only actuall build nightly.

This if statement is already pretty big as is, reducing its complexity makes it more maintainable. If absolutely needed we could always manually run a nigthly.

.github/workflows/build.yml
```diff
@@ -37,7 +114,17 @@ jobs:
           FLUTTER_VERSION=$(jq -r '.flutter' .fvmrc)
           echo "flutter_version=${FLUTTER_VERSION}" >> "$GITHUB_OUTPUT"
 
-          # Print flutter version for logging
+          # Create nightly version with date and time if needed
+          if [[ "${{ needs.check-trigger.outputs.build_type }}" == "nightly" ]]; then
```
PartyDonut, line 109: Best to also use the version code as is, at best we can append the release files inside of the nightly-release with -nightly for clarity sake.

We also don't need the release date here, the only place we actually use it is in the nigtly title, but we could just as well use VERSION_NAME-nightly. Github already adds a date to any release.

.github/workflows/build.yml
```diff
@@ -144,19 +236,21 @@ jobs:
           path: windows/windows_setup.iss
           options: /O+
         env:
-          FLADDER_VERSION: ${{needs.fetch-info.outputs.version_name}}
+          FLADDER_VERSION: ${{ needs.fetch-info.outputs.build_type == 'nightly' && needs.fetch-info.outputs.nightly_version || needs.fetch-info.outputs.version_name }}
```
PartyDonut: Always using version_name would also simplify this logic.

.github/workflows/build.yml
```diff
@@ -279,10 +375,12 @@ jobs:
         with:
           name: fladder-linux
           path: build/linux/x64/release/bundle
+          retention-days: ${{ needs.fetch-info.outputs.build_type == 'nightly' && 7 || 90 }}
 
       - name: Build AppImage
         run: |
-          sed -i -E 's/(version:\s*)latest/\1${{needs.fetch-info.outputs.version_name}}/' AppImageBuilder.yml
+          VERSION_TO_USE="${{ needs.fetch-info.outputs.build_type == 'nightly' && needs.fetch-info.outputs.nightly_version || needs.fetch-info.outputs.version_name }}"
```
PartyDonut: Same here just version_name

.github/workflows/build.yml
```diff
@@ -490,12 +591,206 @@ jobs:
             Fladder-Linux-${{needs.fetch-info.outputs.version_name}}.AppImage
             Fladder-Linux-${{needs.fetch-info.outputs.version_name}}.AppImage.zsync
 
+  create_nightly_release:
```
PartyDonut: Instead of a separate nightly release we could just use the original release step. I would suggest merging them and setting the prerelease toggle according to the build_type

.github/workflows/build.yml
```diff
@@ -144,19 +236,21 @@ jobs:
...
         env:
-          FLADDER_VERSION: ${{needs.fetch-info.outputs.version_name}}
+          FLADDER_VERSION: ${{ needs.fetch-info.outputs.build_type == 'nightly' && needs.fetch-info.outputs.nightly_version || needs.fetch-info.outputs.version_name }}
 
       - name: Archive Windows portable artifact
         uses: actions/upload-artifact@v4.0.0
         with:
           name: fladder-windows-portable
           path: build\windows\x64\runner\Release\
+          retention-days: ${{ needs.fetch-info.outputs.build_type == 'nightly' && 7 || 90 }}
```
PartyDonut: The current max retention is 12 these toggles can probably be removed.

.github/workflows/build.yml
```diff
@@ -490,12 +591,206 @@ jobs:
...
+            Fladder-iOS-${{needs.fetch-info.outputs.nightly_version}}.ipa
+            Fladder-macOS-${{needs.fetch-info.outputs.nightly_version}}.dmg
+            Fladder-Web-${{needs.fetch-info.outputs.nightly_version}}.zip
+            Fladder-Linux-${{needs.fetch-info.outputs.nightly_version}}.zip
+            Fladder-Linux-${{needs.fetch-info.outputs.nightly_version}}.AppImage
+            Fladder-Linux-${{needs.fetch-info.outputs.nightly_version}}.AppImage.zsync
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Clean up old workflow artifacts
```
PartyDonut: Artifacts are only retained for 12 days, so probably don't need this.

PartyDonut (changes requested):
Thanks for merging the build.yml left some comments, mostly for simplifying it a bit so it becomes easier to maintain.

.github/workflows/build.yml
```diff
@@ -379,8 +460,87 @@ jobs:
...
+      - name: Checkout repository
+        if: needs.fetch-info.outputs.build_type == 'nightly'
+        uses: actions/checkout@v4.1.1
+        with:
+          fetch-depth: 0
+
+      - name: Set version variables
+        id: version
+        run: |
+          if [[ "${{ needs.fetch-info.outputs.build_type }}" == "nightly" ]]; then
```
PartyDonut, line 477: This should probably also account for this pattern

echo "BUILD_MODE=profile" >> $GITHUB_ENV
echo "ARTIFACT_SUFFIX=production" >> $GITHUB_ENV
echo "AAB_PATH=productionProfile" >> $GITHUB_ENV
 
We could also export the artifact_suffix variable. Considering create-release is already dependend on build-android, that way we avoid any confusion.

lib/util/update_checker.dart
```diff
@@ -147,4 +147,4 @@ class UpdateChecker {
     }
     return 0;
   }
-}
+}
```
PartyDonut: Missing newline, small nitpick, but very easy fix 😃

Issue #101, opened 2024-11-01
🐛 Android client intiates sync - then nothing happens


Viceman256:
### Description

Everytime I try to sync an item, it starts to sync, then is in an orange status and never syncs anything. If I hit the cloud icon, it spins for 30 seconds then stops. Perhaps a permission issue? The app didn't request file storage access and there's no option to give it.

### Reproduction steps

1. Click Sync
2. Says sync initiated
3. Never syncs

### Screenshots

![DESCRIPTION](LINK.png)

### Platform

Android

### App Version

0.3.1 Build72

PartyDonut:
Hi 

The application uses the internal apps storage so it shouldn't need any storage permission.

One thing missing is the notifications permission, the downloader also shows a android notification. Could you try enabling that and check whether it creates a download task. 
Leaving the notifications disabled should not stop it from downloading but it does give me some more info on what is happening.

If you could also provide information on the media you're trying to sync that would be great.

Viceman256:
I don't think my email sent. 

Hi, thanks for your quick response!

I did enable notifications but unfortunately nothing changed. I just reinstalled it and wanted to reconfigure it with notifications enabled first, but it performed the same. 

I've attached a video to this email response. A screen recording. Did it attach correctly? Basically if you can see, it changes to partial status immediately and stays there. The series don't say partial but they are in the same status. 

https://imgur.com/a/acVNAqq

PartyDonut:
Strange, not sure it seems like it's not even attempting to start the download task itself. 

You do seem to be using a vpn connection it seems? The downloads at the moment are locked to requiring wi-fi that could be the reason it's not downloading anything.
Can you try downloading it using a Wi-fi connection without vpn and also make sure android doesn't think of it as a metered connection.

No logging at the moment unfortunately. 
Another thing you could try if possible, download the application for desktop and try downloading over there. If that works we can atleast pinpoint the issue to mobile.

Viceman256:
It does perform the same on wifi without VPN, but I just tried on the desktop app per your suggestion and experienced the same behavior!

I went into Jellyfin logs and found this everytime I initiate a sync:

[2024-11-04 00:06:16.185 -05:00] [INF] [6] Jellyfin.Api.Helpers.MediaInfoHelper: User policy for "Sibyl". EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
[...]

PR #378, opened 2025-06-08
Update sync_provider.dart

Sopro2015:
## Pull Request Description

Refactored sync_provider.dart to improve download functionality and code quality:

- Used URI helpers for safer URL construction, avoiding manual string concatenation issues.
- Integrated the centralized `dioProvider` for HTTP requests.
- Added Dio interceptors to inject authentication tokens via headers.
- Enhanced error handling by specifically catching DioError exceptions.
- Migrated download progress reporting to use `downloadProgressProvider`.
- Improved logging to support localization and avoid console prints.

## Issue Being Fixed

This addresses issues with unstable download URLs, inconsistent token usage, missing download progress in UI, and unstructured error handling.

## Screenshots / Recordings

(Optional) Screenshots of the improved download process and progress UI.

## Checklist

- [x] Verified correct URL construction and API calls.
- [x] Confirmed progress updates appear in the UI.
- [x] Error handling improved and tested.

### DIFF
--- a/lib/providers/sync_provider.dart
+++ b/lib/providers/sync_provider.dart
@@ -3,6 +3,9 @@ import 'dart:convert';
 import 'dart:developer';
 import 'dart:io';
 
+import 'package:dio/dio.dart';
+import 'dart:io';
+
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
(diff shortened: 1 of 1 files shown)


PartyDonut:
Hi, thanks for looking at fixing this and I appreciate the enthusiasm. However I'm not sure how this fixes the issue, and if it doesn't include issues in all the other platforms.

My first question would be what was the actual issue causing the background downloader to fail? The Android downloads did work for most people so there must be something specific to those few. Replacing the entire download process to fix an unknown might not be the best course of action.

The only real change in the http request itself seems to be that you added the api key to the header of each download. would this not also fix the downloads for the background downloader?

The background downloader was also chosen for its implementation of queuing, pausing and things like notifications or max concurrent downloads, this seems to not implement those things?

Including dio also adds another package while we already have both chopper and the background downloader to handle HTTP requests.

Unrelated to the implementation but its recommended to keep PR's that fix one issue together. This makes it easier for me to review the PR. 
On that note I will be closing the PR's for now. We can discuss further in the issue itself #101

PR #383, opened 2025-06-09
fix: Locale selection issue with country-specific locales

Uncle-Tio:
## Pull Request Description
Fixed locale selection issue where country-specific locales (like `pt_BR`) were incorrectly falling back to language-only locales (like `pt`). Also improved the UI to display both language and country codes in the locale selector.

## Issue Being Fixed
When users selected `pt_BR` in the interface, the app would use `pt` instead of _Brazilian Portuguese_ because the matching logic only compared languageCode, ignoring countryCode.

## Changes Made
1. **Locale matching logic:**
   - Primary match: Exact match for both language and country codes
   - Fallback match: Language-only match if exact match fails
   - Default fallback _(unchanged)_: English (GB) if no matches found 

2. **Locale display in popup menu:**
   - Now shows country codes alongside language codes
   - Format: "Native Name (LANG-COUNTRY)" or "Native Name (LANG)" for language-only locales

## Checklist

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -291,8 +291,11 @@ class _MainState extends ConsumerState<Main> with WindowListener, WidgetsBinding
             builder: (context, child) => Localizations.override(
               context: context,
               locale: AppLocalizations.supportedLocales.firstWhere(
-                (element) => element.languageCode == language.languageCode,
-                orElse: () => const Locale('en', "GB"),
+                (element) => element.languageCode == language.languageCode && element.countryCode == language.countryCode,
+                orElse: () => AppLocalizations.supportedLocales.firstWhere(
+                  (element) => element.languageCode == language.languageCode,
+                  orElse: () => const Locale('en', "GB"),
+                ),
               ),
               child: LocalizationContextWrapper(child: ScaffoldMessenger(child: child ?? Container())),
             ),
--- a/lib/screens/settings/client_sections/client_settings_visual.dart
+++ b/lib/screens/settings/client_sections/client_settings_visual.dart
@@ -46,7 +46,7 @@ List<Widget> buildClientSettingsVisual(
                         context: context,
                         locale: entry,
                         child: Builder(builder: (context) {
-                          return Text("${context.localized.nativeName} (${entry.languageCode.toUpperCase()})");
+                          return Text("${context.localized.nativeName} (${entry.languageCode.toUpperCase()}${entry.countryCode != null ? '-${entry.countryCode!.toUpperCase()}' : ''})");
                         }),
                       ),
                       onTap: () => ref


lib/main.dart
```diff
@@ -291,8 +291,11 @@ class _MainState extends ConsumerState<Main> with WindowListener, WidgetsBinding
             builder: (context, child) => Localizations.override(
               context: context,
               locale: AppLocalizations.supportedLocales.firstWhere(
-                (element) => element.languageCode == language.languageCode,
-                orElse: () => const Locale('en', "GB"),
+                (element) => element.languageCode == language.languageCode && element.countryCode == language.countryCode,
```
PartyDonut: Most other languages only have the languageCode itself not the country code.

A better solution would be to first fetch all languages with the languageCode and then check if any of those languages also have the countryCode. And either fallback to the first available?

For instance these are the current language codes.

`[ar, cs, de, en, es, fr, it, ja, ko, mr, mt, nb, nl, pl, pt, pt_BR, ro, ru, sk, ta, tr, uk, zh]`

lib/screens/settings/client_sections/client_settings_visual.dart
```diff
@@ -46,7 +46,7 @@ List<Widget> buildClientSettingsVisual(
                         context: context,
                         locale: entry,
                         child: Builder(builder: (context) {
-                          return Text("${context.localized.nativeName} (${entry.languageCode.toUpperCase()})");
+                          return Text("${context.localized.nativeName} (${entry.languageCode.toUpperCase()}${entry.countryCode != null ? '-${entry.countryCode!.toUpperCase()}' : ''})");
```
PartyDonut: Lets move the last part `(${entry.languageCode.toUpperCase()}${entry.countryCode != null ? '-${entry.countryCode!.toUpperCase()}' : ''})"` to a extension method.

PartyDonut (changes requested):
Thanks for fixing this I left a few comments
Let's also move both comments functionality to a locale_extension.dart, currently all extensions are located at lib/util.

lib/main.dart
Uncle-Tio: I don't know if I understood correctly, the current implementation prioritizes the most specific locale first, then falls back to the general one if unavailable. I believe this approach is generally better, prioritizing country-specific locales is preferable due to significant regional differences.

For example:
- `pt_BR` vs `pt` have different vocabulary, currency formats, and cultural contexts
- `en_US` vs `en_GB` differ in spelling, date formats, and terminology
- `es_MX` vs `es_ES` have distinct expressions and cultural references

I understand that only one language has countryCode at the moment, but wouldn't it be more future proof?

Also I didn't take into account how Flutter/Fladder automatically selects the language (first boot). In my case, it was correctly selecting pt_BR, and none of the other languages were affected by this change (the translation itself). But it's necessary to test if the (automatic) selection is okay in the other languages...

lib/screens/settings/client_sections/client_settings_visual.dart
Uncle-Tio: Maybe put it in `localization_helper.dart`?

PartyDonut: That would also be fine.

lib/main.dart
PartyDonut: It would be more future proof, but we still need the check without countryCode. For instance dutch only has the langugae code (nl). But there is Belgium who use Dutch (nl_BE). 
However there is no nl_NL in weblate despite most systems returning nl_NL as locale.

Requiring the countryCode at this point would just break all languages except for pt_BR. So for now I think the safest bet is to add the countryCode check.

In short we should first select all items with the same languageCode and then check for countryCode. If there is no translation for that specific countryCode we should fall-back to the language that has no countryCode. And if there is no translation with that langaugeCode fall back to English.

At this point there are not enough translations to warrent such a specific locale(countryCode) check.

Uncle-Tio: I believe that's what I've done:

1. **It first tries to match `languageCode + countryCode`:**

   ```dart
   firstWhere(
     (e) => e.languageCode == language.languageCode
         && e.countryCode  == language.countryCode,
     …
   )
   ```
2. **If none is found**, it falls back to matching only the `languageCode`:

   ```dart
   orElse: () => AppLocalizations.supportedLocales.firstWhere(
     (e) => e.languageCode == language.languageCode,
     orElse: () => const Locale('en', 'GB'),
   ),
   ```
3. **And finally**, if there’s still no match, it falls back to English (`en_GB`).

To ensure that the “language-only” fallback picks **only** the locale without a country, we could refine the second search like this:

```dart
locale: AppLocalizations.supportedLocales.firstWhere(
  (e) => e.languageCode == language.languageCode
      && e.countryCode == language.countryCode,
  orElse: () => AppLocalizations.supportedLocales.firstWhere(
    (e) => e.languageCode == language.languageCode
        && (e.countryCode == null || e.countryCode!.isEmpty),
    orElse: () => const Locale('en', 'GB'),
  ),
),
```

This guarantees that the `languageCode` fallback won’t accidentally return a locale with a different `countryCode`.

This is actually the default behavior of the MaterialApp:: https://api.flutter.dev/flutter/material/MaterialApp/supportedLocales.html
In fact, if we do it this way, we seem to get the same result (if you want to simplify):
```
            localizationsDelegates: AppLocalizations.localizationsDelegates,
            supportedLocales: AppLocalizations.supportedLocales,
            locale: language,
            builder: (context, child) => LocalizationContextWrapper(
              child: ScaffoldMessenger(child: child ?? Container()),
            ),
```

Uncle-Tio:
Take a look at the changes and see what you think. A few points:

- Now using the default behavior of MaterialApp ([https://api.flutter.dev/flutter/material/MaterialApp/supportedLocales.html](https://api.flutter.dev/flutter/material/MaterialApp/supportedLocales.html)):

   1. Locale.languageCode, Locale.scriptCode, and Locale.countryCode
   2. Locale.languageCode and Locale.scriptCode only
   3. Locale.languageCode and Locale.countryCode only
   4. Locale.languageCode only
   5. Locale.countryCode only when all preferred locales fail to match
   6. Returns the first element of supportedLocales as a fallback

   However, I changed step 6 to use English as the final fallback.
   Also we're not using `Locale.scriptCode`, but that doesn’t seem to be an issue.

- Moved it to an extension in `localization_helper`, as you requested.

- Changed "Unknown" to "English" because when it fell back to English, it was displaying "Unknown"; 
I don't think that's a problem? Languages without a `nativeName` were already returning "English" for some reason, but I'm not sure if we'll actually run into cases with no `nativeName`. The cases I noticed were due to the language file being empty [(app_tr.arb](https://github.com/DonutWare/Fladder/blob/develop/lib/l10n/app_tr.arb), might be worth checking that.
(But if you want to find out why it says Unknown, we can sort it out first)

Let me know if anything's off. Sorry for not opening a discussion before the pull request, I thought it would be a small change. I'll update the initial description as well.

# Testing
- [x] Manually tested the app (Edge) by setting the system locale to an unsupported language: fallback correctly selected English as expected.
- [x] Verified that supported locales such as Korean (ko) and Brazilian Portuguese (pt-BR) were automatically detected and applied correctly.

PR #391, opened 2025-06-15
fix: big files downloads failing in flatpak app

lmgarret:
## Pull Request Description

See https://github.com/DonutWare/Fladder/discussions/381 for context. The flatpak build of the app was writing temporary files in a in-memory `/tmp` directory (as provided by flatpak by default). This was causing downloads of bigger files to fail as it was filling this tmp dir. This PR fixes it by using the host machine's `/tmp` directory.

While I was at it, I also upgraded to use a newer platform  which fixes retries on dowload being stuck, and simplified the flatpak build, which now takes ~5m compared to ~14m before.

## Issue Being Fixed

 - downloads often failing in flatpak
 - retrying downloads stuck (side effect of ugprade the platform used)
Resolves #issue-number

## Screenshots / Recordings

NA

## Checklist


PartyDonut (approved):
lgtm, thanks for fixing this.

Issue #403, opened 2025-06-30
🐛 Stop subtitle from going to middle of screen


vikingnope:
### Description

Currently when the menu is being displayed in the player the subtitles go into the middle of the screen as per the screenshot, which makes the viewing impossible to see.

### Reproduction steps

```bash
1. Start playing any video
2. Make the menu show with the subtitles on and they go into the middle
```

### Screenshots

![Image](https://github.com/user-attachments/assets/175d613f-36ad-4a62-acd5-8f0edf479048)

I would suggest putting it somewhere here:

![Image](https://github.com/user-attachments/assets/00f1d9d3-a21a-4f1d-9737-5080d6418862)

### Logs

```bash

```

### Platform

Android, Linux, Windows, Macos

### App Version

0.7.0

PR #420, opened 2025-07-27
fix: Subtitle Positioning Issue During Menu Display

bk-bf:
## 📋 Summary

Fixed subtitle positioning bug where subtitles jumped to screen center when video player menu was displayed. Now uses dynamic menu height measurement for precise subtitle positioning just above the menu.

## 🐛 Problem
Subtitles rendered in the middle of the screen when the video player menu overlay was active, obscuring video content and making it annoying to watch.

### Key Changes
- **Dynamic height measurement**: Real-time calculation of actual menu dimensions
- **Intelligent positioning**: Places subtitles just above menu with collision detection
- **Smart fallback**: Uses 15% screen height when measurement unavailable
- **Respects user preferences**: Only adjusts subtitles that would overlap with menu

### Modified Files
- `lib_mpv.dart` - Dynamic subtitle positioning logic
- `video_player_controls.dart` - Menu height measurement
- `base_player.dart` - Updated interface for menuHeight parameter
- `lib_mdk.dart` - API compatibility updates

## ✅ Benefits
- **Pixel-perfect positioning** using actual menu dimensions
- **Minimal visual disruption** with subtitles positioned just above menu
- **Maintains user preferences** for high-positioned subtitles
- **Cross-platform consistency** with graceful fallback behavior

## 🧪 Testing
Verified on macOS with various scenarios including dynamic height calculation, menu toggling, high-positioned subtitles, fallback behavior, and edge cases.

## 🔄 Backwards Compatibility
✅ Fully compatible - no breaking changes, preserves existing subtitle behavior and user preferences.

**Resolves:** https://github.com/DonutWare/Fladder/issues/403

### DIFF
--- a/lib/wrappers/players/lib_mpv.dart
+++ b/lib/wrappers/players/lib_mpv.dart
@@ -168,13 +168,13 @@ class LibMPV extends BasePlayer {
   @override
   Widget? subtitles(
     bool showOverlay, {
-    GlobalKey? menuKey,
+    GlobalKey? controlsKey,
   }) =>
       _controller != null
           ? _VideoSubtitles(
               controller: _controller!,
               showOverlay: showOverlay,
-              menuKey: menuKey,
+              controlsKey: controlsKey,
             )
           : null;
 
@@ -198,12 +198,12 @@ class LibMPV extends BasePlayer {
 class _VideoSubtitles extends ConsumerStatefulWidget {
   final VideoController controller;
   final bool showOverlay;
-  final GlobalKey? menuKey;
+  final GlobalKey? controlsKey;
 
   const _VideoSubtitles({
     required this.controller,
     this.showOverlay = false,
-    this.menuKey,
+    this.controlsKey,
   });
 
   @override
@@ -275,9 +275,9 @@ class _VideoSubtitlesState extends ConsumerState<_VideoSubtitles> {
 
   void _measureMenuHeight() {
     WidgetsBinding.instance.addPostFrameCallback((_) {
-      if (!mounted || widget.menuKey == null) return;
+      if (!mounted || widget.controlsKey == null) return;
 
-      final RenderBox? renderBox = widget.menuKey?.currentContext?.findRenderObject() as RenderBox?;
+      final RenderBox? renderBox = widget.controlsKey?.currentContext?.findRenderObject() as RenderBox?;
       final newHeight = renderBox?.size.height;
 
       if (newHeight != _cachedMenuHeight && newHeight != null) {
--- a/lib/stubs/web/lib_mdk_web.dart
+++ b/lib/stubs/web/lib_mdk_web.dart
@@ -64,7 +64,7 @@ class LibMDK extends BasePlayer {
       null;
 
   @override
-  Widget? subtitles(bool showOverlay, {GlobalKey? menuKey}) => null;
+  Widget? subtitles(bool showOverlay, {GlobalKey? controlsKey}) => null;
 
   @override
   Future<void> setVolume(double volume) async {}
--- a/lib/wrappers/media_control_wrapper.dart
+++ b/lib/wrappers/media_control_wrapper.dart
@@ -40,7 +40,7 @@ class MediaControlsWrapper extends BaseAudioHandler {
   PlayerState? get lastState => _player?.lastState;
 
   Widget? subtitleWidget(bool showOverlay, {GlobalKey? controlsKey}) =>
-      _player?.subtitles(showOverlay, menuKey: controlsKey);
+      _player?.subtitles(showOverlay, controlsKey: controlsKey);
   Widget? videoWidget(Key key, BoxFit fit) => _player?.videoWidget(key, fit);
 
   final Ref ref;
--- a/lib/wrappers/players/base_player.dart
+++ b/lib/wrappers/players/base_player.dart
@@ -20,7 +20,7 @@ abstract class BasePlayer {
   );
   Widget? subtitles(
     bool showOverlay, {
-    GlobalKey? menuKey,
+    GlobalKey? controlsKey,
   });
   Future<void> dispose();
   Future<void> open(String url, bool play);
--- a/lib/wrappers/players/lib_mdk.dart
+++ b/lib/wrappers/players/lib_mdk.dart
@@ -191,7 +191,7 @@ class LibMDK extends BasePlayer {
             );
 
   @override
-  Widget? subtitles(bool showOverlay, {GlobalKey? menuKey}) => null;
+  Widget? subtitles(bool showOverlay, {GlobalKey? controlsKey}) => null;
 
   @override
   Future<void> setVolume(double volume) async => _controller?.setVolume(volume / 100);


bk-bf:
@sourcery-ai review

bk-bf:
sourcery-ai review

bk-bf:
@sourcery-ai review

lib/wrappers/players/lib_mpv.dart
```diff
@@ -224,24 +237,54 @@ class _VideoSubtitlesState extends ConsumerState<_VideoSubtitles> {
     super.dispose();
   }
```
SourceryAI, line 240: **suggestion:** Subtitle offset calculation now considers menu height, but may not handle rapid menu size changes.

If the menu height can change dynamically, ensure the widget rebuilds and recalculates the subtitle offset as needed. Trigger a rebuild on menu height changes if this isn't already handled.

Suggested implementation:

```
    final settings = ref.watch(subtitleSettingsProvider);
    final padding = MediaQuery.of(context).padding;
    final menuHeight = ref.watch(menuHeightProvider);

```

```
    // Process subtitle text
    final text = subtitle.where((line) => line.trim().isNotEmpty).map((line) => line.trim()).join('\n');

    // Use menuHeight in subtitle offset calculation as needed
    // For example, if you pass offset to SubtitleText:
    // final subtitleOffset = calculateSubtitleOffset(menuHeight, ...);

```

- You must ensure that `menuHeightProvider` exists and emits the current menu height. If not, create it and update it whenever the menu height changes.
- Update the subtitle offset calculation and pass the offset to the subtitle rendering widget (e.g., `SubtitleText`) as needed.
- If you use a different state management solution for menu height, adapt the `ref.watch(menuHeightProvider)` line accordingly.

SourceryAI (commented):
Hey @bk-bf - I've reviewed your changes - here's some feedback:

- Consider scheduling a post-frame callback (e.g. WidgetsBinding.addPostFrameCallback) to capture and store the menu height in state so subtitles update immediately once the bottom controls are laid out, preventing stale or null values on initial render.
- The nested math.max(math.min(...)) clamping in _calculateSubtitleOffset can be simplified to a single clamp operation to make the logic more readable.

<details>
<summary>Prompt for AI Agents</summary>

~~~markdown
Please address the comments from this code review:
## Overall Comments
- Consider scheduling a post-frame callback (e.g. WidgetsBinding.addPostFrameCallback) to capture and store the menu height in state so subtitles update immediately once the bottom controls are laid out, preventing stale or null values on initial render.
- The nested math.max(math.min(...)) clamping in _calculateSubtitleOffset can be simplified to a single clamp operation to make the logic more readable.

## Individual Comments

### Comment 1
<location> `lib/wrappers/players/lib_mpv.dart:239` </location>
<code_context>
-    if (widget.controller.player.platform?.configuration.libass ?? false) {
-      return const IgnorePointer(child: SizedBox.shrink());
-    } else {
-      return SubtitleText(
-        subModel: settings,
-        padding: padding,
</code_context>

<issue_to_address>
Subtitle offset calculation now considers menu height, but may not handle rapid menu size changes.

If the menu height can change dynamically, ensure the widget rebuilds and recalculates the subtitle offset as needed. Trigger a rebuild on menu height changes if this isn't already handled.

Suggested implementation:

```
    final settings = ref.watch(subtitleSettingsProvider);
    final padding = MediaQuery.of(context).padding;
    final menuHeight = ref.watch(menuHeightProvider);

```

```
    // Process subtitle text
    final text = subtitle.where((line) => line.trim().isNotEmpty).map((line) => line.trim()).join('\n');

    // Use menuHeight in subtitle offset calculation as needed
    // For example, if you pass offset to SubtitleText:
    // final subtitleOffset = calculateSubtitleOffset(menuHeight, ...);

```

- You must ensure that `menuHeightProvider` exists and emits the current menu height. If not, create it and update it whenever the menu height changes.
- Update the subtitle offset calculation and pass the offset to the subtitle rendering widget (e.g., `SubtitleText`) as needed.
- If you use a different state management solution for menu height, adapt the `ref.watch(menuHeightProvider)` line accordingly.
</issue_to_address>
~~~

</details>

***

Hi @bk-bf! 👋

Thanks for trying out Sourcery by commenting with `@sourcery-ai review`! 🚀

Install the [sourcery-ai bot](https://github.com/apps/sourcery-ai/installations/new) to get automatic code reviews on every pull request ✨

<sub>
Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.
</sub>

lib/util/subtitle_position_calculator.dart
PartyDonut, line 1: I'm more of a less is more person. There are a lot of comments in this code that probably don't need to be there.

test/util/subtitle_position_calculator_test.dart
PartyDonut, line 1: More test are probably better, but not sure if something like this needs a test. Let's remove it.

lib/screens/video_player/video_player_controls.dart
PartyDonut, line 1: Lets move the height calculation to the  _VideoSubtitles class inside of lib_mpv. 
The mdk player for instance doesn't have this functionality but it would still calculate. If we ever have the option for other custom subtitles we can implement different logic there if needed.

PartyDonut (changes requested):
Hi thanks for taking a stab at implementing this.

Also make sure to check if it works on different displays, mobile for example is broken in both portrait and landscape.

Image

Subtitles are overlayed by some UI elements


Also on a unrelated note, the initial comment seems to be at least partially generated by AI?
While it's always good to give as much information as needed I would personally prefer a more to the point note about the implemented changes.

bk-bf:
Hey @PartyDonut , 

I have implemented ur requested changes regarding the comment verbosity and removed the unnecessary test file. As for the height calculation, there is no easy way of migrating it from the video_player_controls.dart file into the _VideoSubtitles class inside of lib_mpv, without drastically changing my implementation logic and it seems more sensible to leave it in there for the following reasons: 

Why measurement should stay in video_player_controls.dart:

- Ownership - Controls widget owns and renders the menu UI
- Access - Only this widget has direct access to the menu's RenderBox
- Separation of concerns - Subtitle widgets shouldn't know about control UI structure
- Reusability - Different players get the same accurate measurement
- Clean architecture - Controls handle UI measurement, players handle positioning

Why video players should receive the measurement:

- Accurate positioning - Subtitle widgets get pre-measured data instead of guessing
- Fallback safety - When null, the calculator uses safe fallback values
- Player independence - Each player type can handle subtitles without measuring external UI

_TL;DR: The height measurement logic should stay in video_player_controls.dart because it owns the menu UI and can directly measure it, while moving it to lib_mpv.dart would require the subtitle widget to awkwardly reach across architectural boundaries to measure UI components it doesn't own._

but maybe i misunderstood ur point, could u give me a clarification on that? Regarding the subtitle positioning issue on mobile, especially in vertical mode, to counteract that i have set the padding to 0 (previously at -0.03 to place the subs closer to the player menu bar), but left it for a possible future implementation of a user-adjustable slider to control subtitle positioning relative to the player menu, the subtitles should now always render above all player menu UI elements, but sadly I don't have a simple way of verifying that on mobile devices, if u don't mind checking if this solution is enough I would be really glad. 

Thanks for ur feedback!

lib/util/subtitle_position_calculator.dart
```diff
@@ -1,12 +1,9 @@
 import 'dart:math' as math;
+
 import 'package:fladder/models/settings/subtitle_settings_model.dart';
 
 class SubtitlePositionCalculator {
   static const double _fallbackMenuHeightPercentage = 0.15;
-  static const double _dynamicSubtitlePadding =
```
PartyDonut: Did not see these at first, but there is already a height offset slider in subtitle customization so we don't need this.

lib/models/settings/subtitle_settings_model.dart
```diff
@@ -216,7 +216,8 @@ class SubtitleText extends ConsumerWidget {
               child: Stack(
                 alignment: Alignment.bottomCenter,
                 children: [
-                  Positioned(
+                  AnimatedPositioned(
```
PartyDonut, line 219: Animated because it's less jarring to have it move in an instant.

bk-bf:
I am absolutely fine with ur changes! I understand now that u wanted a hybrid solution for the height calculation migration, also nice catch with the animation!

PR #424, opened 2025-07-30
fix: Standardize precision used in communityRating

lbux:
## Pull Request Description

I tried to keep this as minimal as possible to only the issue(s) presented in the discussion. The only changes were to the precision of the base model item and the overview header. I've provided screenshots of both changes. Additionally, for the poster widget, I'd like some feedback on how to proceed with the changes. Currently, a ternary operator is used to check if `subTitle` exists to ensure that the year is aligned to the left or right. This has an unintended consequence (image 3) of the year being aligned to the left if an entry does not have a `subTitle` but other entries do. This is inconsistent, but the alternative would be that if at least one item in the whole library filtered/displayed has a `subTitle` then the year must be to the right.

## Issue Being Fixed

Resolves #422 

## Screenshots / Recordings

image

image

image

## Checklist

### DIFF
--- a/lib/models/item_base_model.dart
+++ b/lib/models/item_base_model.dart
@@ -90,7 +90,7 @@ class ItemBaseModel with ItemBaseModelMappable {
                     color: Colors.yellowAccent,
                   ),
                   const SizedBox(width: 6),
-                  Text((overview.communityRating ?? 0.0).toString())
+                  Text((overview.communityRating ?? 0.0).toStringAsFixed(2))
                 ],
               )
             : null,
--- a/lib/models/item_base_model.dart
+++ b/lib/models/item_base_model.dart
@@ -87,7 +87,7 @@ class ItemBaseModel with ItemBaseModelMappable {
                 color: Colors.yellowAccent,
               ),
               const SizedBox(width: 6),
-              Text(overview.communityRating?.toStringAsFixed(2) ?? "0.0"),
+              Text(overview.communityRating?.toStringAsFixed(2) ?? "--"),
             ],
           ),
         _ => null,
--- a/lib/models/item_base_model.dart
+++ b/lib/models/item_base_model.dart
@@ -68,32 +68,28 @@ class ItemBaseModel with ItemBaseModelMappable {
   }
 
   Widget? subTitle(SortingOptions options) => switch (options) {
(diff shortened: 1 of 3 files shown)


lib/models/item_base_model.dart
PartyDonut, line 1: For the problem in image 3. One solution might be to instead of returning null when we have no parental/community-Rating. We return the Row widget.

And if the rating is null we can do this `overview.communityRating?.toStringAsFixed(2) ?? "--"`

lib/screens/shared/media/poster_widget.dart
```diff
@@ -93,12 +93,15 @@ class PosterWidget extends ConsumerWidget {
...
                     if (poster.subText?.isNotEmpty ?? false)
                       Flexible(
-                        child: ClickableText(
-                          opacity: opacity,
-                          text: poster.subText ?? "",
-                          maxLines: 1,
-                          overflow: TextOverflow.ellipsis,
-                          style: Theme.of(context).textTheme.titleSmall?.copyWith(fontWeight: FontWeight.bold),
+                        child: Align(
+                          alignment: subTitle == null ? Alignment.centerLeft : Alignment.centerRight,
```
PartyDonut: Using alignment here is a bit strange. Would be better to wrap the Opacity widget on line 88 with a Flexible as well.

Then for the row you can set the following variable to this.
`mainAxisAlignment: MainAxisAlignment.spaceBetween,
`

lib/models/item_base_model.dart
lbux: I've implemented these changes. However, I made "--" for parentalRating and 0.0 for communityRating since those make more sense imo. But I can make both "--" if preferred.

lib/screens/shared/media/poster_widget.dart
lbux: I added the flexible but removed the spacer because it would cause a renderflex overflow. Because of the mainAxisAlignment, it isnt necessary anymore

lbux:
Despite not changing anything to do with gradle, I can't get debugging to work using VsCode. I manually moved the generated .apk to my phone and tested it. No issues with UI.

PartyDonut (approved):
Looks good to me, thanks for implementing this 😄

PR #430, opened 2025-08-03
fix: make image cache key unique

lbux:
## Pull Request Description

The current implementation assumes that Jellyfin's tags are unique and can therefore be used as keys. However, [their logic to create tags](https://github.com/jellyfin/jellyfin/blob/9e36fa426335289b4be90034640ad38b38e3de79/src/Jellyfin.Drawing/ImageProcessor.cs#L409) can lead to duplicate tags across images in the same location (backdrop.jpg, logo.jpg, etc can share the same tag) and same timestamp.

Consider this url example from my server. The tags are the same:

http://localhost:8096/Items/61139a4c4ce0f3e26032c35a7e1adde8/Images/Primary?maxWidth=575&tag=497955eaadd4d4d31609987a736d8374

http://server:8096/Items/61139a4c4ce0f3e26032c35a7e1adde8/Images/Logo?maxWidth=575&tag=497955eaadd4d4d31609987a736d8374

Since the key is assumed to be unique, when we request a logo, we do so with the key. However, this often leads to a primary or backdrop being returned since in our cache, that key belongs to one of those. See below for screenshots.

## Issue Being Fixed

N/A

## Screenshots / Recordings

Fladder 0.7.0:

fladder_xZQthz1iw6
fladder_dq6Rsbak4n

This PR:
fladder_dRdGlrFNgH
fladder_yH4EeruV4A

## Checklist

### DIFF
--- a/lib/models/item_base_model.dart
+++ b/lib/models/item_base_model.dart
@@ -87,7 +87,11 @@ class ItemBaseModel with ItemBaseModelMappable {
                 color: Colors.yellowAccent,
               ),
               const SizedBox(width: 6),
+<<<<<<< HEAD
               Text(overview.communityRating?.toStringAsFixed(2) ?? "--"),
+=======
+              Text(overview.communityRating?.toStringAsFixed(2) ?? "0.0"),
+>>>>>>> 9088937 (implement feedback)
             ],
           ),
         _ => null,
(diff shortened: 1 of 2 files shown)


lbux:
I just realized I included the history from my previous PR. I tried to clean it up. Hopefully it works lol

lbux:
@sourcery-ai dismiss

lbux:
@sourcery-ai review

PR #438, opened 2025-08-06
feat: adds JKLM controls

ynot01:
## Pull Request Description

Adds J key as an alternative to arrowLeft (Seek back)

Adds K key as an alternative to space (Play/Pause)

Adds L key as an alternative to arrowRight (Seek forward)

Adds M key as mute, which sets volume to 0, identical to the clickable mute button

## Issue Being Fixed

Resolves the first 4 of #433

## Screenshots / Recordings

No visual changes

## Checklist

### DIFF
--- a/lib/screens/video_player/components/video_player_seek_indicator.dart
+++ b/lib/screens/video_player/components/video_player_seek_indicator.dart
@@ -50,21 +50,21 @@ class _VideoPlayerSeekIndicatorState extends ConsumerState<VideoPlayerSeekIndica
 
   bool _onKey(KeyEvent value) {
     if (value is KeyRepeatEvent) {
-      if (value.logicalKey == LogicalKeyboardKey.arrowLeft) {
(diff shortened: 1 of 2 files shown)


lbux:
I think the state of the current volume setting should be saved to later restore. On Youtube, when you press m again, it will restore it to the previous volume %. From the code it doesn't seem to do that currently, correct?

ynot01:
> I think the state of the current volume setting should be saved to later restore. On Youtube, when you press m again, it will restore it to the previous volume %. From the code it doesn't seem to do that currently, correct?

Yes, identical to the clickable mute button

PartyDonut:
Hi, thanks for taking your time to implement this. 

While the code does look fine and correct, I already started on implementing a customizable hotkey system for Fladder. This replaces/refactors the code you changed in this PR.

It's always a good idea to check the "I would like to implement this myself!" checkbox, even if you want to partially implement it. So we can avoid having multiple working on the same feature.

Implemented in #439, closing this one.

ynot01:
No problem at all

PR #448, opened 2025-08-11
chore: Winget package automation

vikingnope:
## Pull Request Description

Adds winget package automation so that winget packages are updated automatically with each release.

For this to work a fork of the https://github.com/microsoft/winget-pkgs repo needs to be done on the DonutWare account, and then a classic token needs to be created (make sure to set it to never expire, unless you want to change it every couple of months) and added to this repo in the secrets section and it needs to be called `WINGET_TOKEN`.

I also decided to keep it to a specific SHA because the latest version v2 is not up to date with the newest changes which are needed for smooth operation.

More info can be found here: https://github.com/vedantmgoyal9/winget-releaser

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -716,3 +716,30 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./build/web
+
+  # Winget package submission for release builds
+  winget-package-submit-release:
+    runs-on: ubuntu-latest
+    needs: [check-trigger, create_release]
+    if: needs.check-trigger.outputs.build_type == 'release'
+    steps:
+      - name: Submit to Winget
+        uses: vedantmgoyal9/winget-releaser@19e706d4c9121098010096f9c495a70a7518b30f
+        with:
+          identifier: DonutWare.Fladder
+          installers-regex: 'Fladder-Windows-.*-Setup\.exe$'
+          token: ${{ secrets.WINGET_TOKEN }}
+
+  # Winget package submission for nightly builds
+  winget-package-submit-nightly:
+    runs-on: ubuntu-latest
+    needs: [check-trigger, create_release]
+    if: needs.check-trigger.outputs.build_type == 'nightly'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -716,3 +716,26 @@ jobs:
...
+  winget-package-submit:
+    runs-on: ubuntu-latest
+    needs: [check-trigger, create_release]
+    strategy:
+      matrix:
+        include:
+          - build_type: release
+            identifier: DonutWare.Fladder
+            max_versions: ""
+          - build_type: nightly
```
PartyDonut, line 730: Would nightly builds not be annoying for them to review. Atleast I'm assuming these need to be reviewed every time a build is created?

Also probably not a lot of interest in nightly builds for winget users.

.github/workflows/build.yml
```diff
@@ -716,3 +716,26 @@ jobs:
...
+            identifier: DonutWare.Fladder.Nightly
+            max_versions: "1"
+    steps:
+      - name: Submit to Winget (${{ matrix.build_type }})
+        if: needs.check-trigger.outputs.build_type == matrix.build_type
+        uses: vedantmgoyal9/winget-releaser@19e706d4c9121098010096f9c495a70a7518b30f
```
PartyDonut, line 736: What problems was v2 giving? It's always preferred to use the latest stable version.

schembriaiden: I believe the v2 does not have the following fixes:

image


The winget sync fork is the most important one as there was some change to the GitHub API and the Komac repo has not gotten to fixing it yet.

.github/workflows/build.yml
schembriaiden: I thought since it is available we could add it. It doesn't really matter with the annoying part as there are a bunch of other packages which also have nightly builds on winget.

PartyDonut (approved):

PartyDonut:
Created the fork and added the WINGET_TOKEN secret 👍🏼 .

PR #468, opened 2025-08-19
ui: Change icon letter to uppercase

vikingnope:
## Pull Request Description

It is a small design change that I think looks better in my opinion. Now the initial on the icon is uppercase, even when the username starts with a lowercase.

## Screenshots / Recordings

Before:

image

After:

image

## Checklist


PartyDonut:
I had the same idea at first, the reason I did not implement it was because this forces a certain casing. There might be users who actually want it to be lowercase. So my suggestion for users would be make the user account name capitalized.

PR #484, opened 2025-09-07
Feat speed shortcuts 

Julien9969:
Introduces playback speed control shortcuts and speed indicator overlay, and the necessary localization and state management updates.

**Playback Speed Control Improvements:**

* Added `speedUp` and `speedDown` hotkeys to the `VideoHotKeys` enum, updated their display names, serialization, and default key combinations (Ctrl+Up/Ctrl+Down). (`lib/models/settings/video_player_settings.dart`, `lib/models/settings/video_player_settings.g.dart`) 

* Implemented handling of the new speed hotkeys in the desktop video player controls, calling the new `steppedSpeed` method to adjust playback rate. (`lib/screens/video_player/video_player_controls.dart`)

**Playback Speed State and Indicator:**

* Introduced a `playbackRateProvider` using Riverpod for managing the current playback speed, and a `steppedSpeed` method in the settings notifier to update the speed, clamp it, and notify the player. (`lib/providers/settings/video_player_settings_provider.dart`) 
* Added a new `VideoPlayerSpeedIndicator` widget that listens for speed changes and displays an overlay with the current playback rate and an icon, shown briefly after each adjustment. (`lib/screens/video_player/components/video_player_speed_indicator.dart`, `lib/screens/video_player/video_player_controls.dart`)

## Screenshots / Recordings
image

[Screencast_20250907_153343.webm](https://github.com/user-attachments/assets/5dd842d9-f56d-479c-968b-129daf90f405)

## Checklist

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -1302,6 +1302,8 @@
   "mute": "Mute",
   "volumeUp": "Volume Up",
   "volumeDown": "Volume Down",
+  "speedUp": "Speed Up",
+  "speedDown": "Speed Down",
   "nextVideo": "Next Video",
   "prevVideo": "Previous Video",
   "nextChapter": "Next Chapter",
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -1327,5 +1327,13 @@
         "type": "int"
       }
     }
+  },
+  "speedIndicator": "Playback rate: {speed}",
+  "@speedIndicator": {
+    "placeholders": {
+      "speed": {
+        "type": "double"
+      }
+    }
   }
 }
--- a/lib/models/settings/video_player_settings.dart
+++ b/lib/models/settings/video_player_settings.dart
@@ -20,6 +20,8 @@ enum VideoHotKeys {
   mute,
   volumeUp,
   volumeDown,
+  speedUp,
+  speedDown,
   nextVideo,
   prevVideo,
   nextChapter,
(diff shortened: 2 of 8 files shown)


schembriaiden:
Hiya, as a suggestion, I thought it would be nicer if the shortcut for this is changed to the same one as youtube as it is quite nice imo.

So for faster playback: SHIFT and > (together)
and for slower playback: SHIFT and < (together)

Julien9969:
> Hiya, as a suggestion, I thought it would be nicer if the shortcut for this is changed to the same one as youtube as it is quite nice imo.
> 
> So for faster playback: SHIFT and > (together) and for slower playback: SHIFT and < (together)

I choose this because it's the shortcut defined in MPC.
I'm open to something else but with azerty keyboard, the < and > are on the same key, commas, dot etc are not at the same place so the shortcuts for this will be messy.
I think it's better to choose key combinations that are common to all keyboards. (like crtl + up)

schembriaiden:
> I choose this because it's the shortcut defined in MPC. I'm open to something else but with azerty keyboard, the < and > are on the same key, commas, dot etc are not at the same place so the shortcuts for this will be messy. I think it's better to choose key combinations that are common to all keyboards. (like crtl + up)

Oh ok makes, sense, then just leave them as is :). Thanks for the consideration.

PartyDonut (changes requested):
While the code itself looks good. There is now a small inconsistency.

Because the step size for the hotkey is .10 and the minimum is 0.25 we need to go all the way up to 3.00 to be able to go back to a nice 1.00 (original) playback rate.

The reason for this is that the playbackRate slider in the extra options in the bottom left has a stepSize of 0.05.

Setting the hotkey stepSize to 0.05 would work. Or we could have the "last" step be 0.05 and all the other steps stay 0.1. The only thing that's important is being able to go back to the original 1x play rate quickly

Julien9969:
```
  void steppedSpeed(double i) {
    var value = double.parse(
      ((ref.read(playbackRateProvider) + i).clamp(0.25, 3))
          .toStringAsFixed(2),
    );

    if ((value - 1.0).abs() < 0.1) {
      value = 1.0;
    }
    ref.read(playbackRateProvider.notifier).state = value;
    ref.read(videoPlayerProvider).setSpeed(value);
  }
  ```
  
  If the speed 0.95 and 1.05 the speed will be forced to 1

PartyDonut (approved):

PartyDonut (approved):
Thanks for implementing this. 

Small fix because the check gets locked in the 1-3 playback rate range. When checking for absolute of < 0.1.

PR #508, opened 2025-09-30
feat: Add support for taking screenshots with hotkeys

RainbowCookie32:
## Pull Request Description
My take on implementing my request on https://github.com/DonutWare/Fladder/discussions/497. This bring the possibility to take screenshots with or without subtitles using hotkeys. The screenshots are saved to a folder picked by the user, using an incrementing number for file name, which can be optionally zero-padded (by default, it's padded to 3 digits).

The feature works on both backends on desktop and (I think) should only show up on settings on that platform. The default keybind is G for screenshots with subtitles, and Ctrl+G for screenshots without subtitles. Gave it 0 thought though and I was trying to figure out inputs and how hotkeys worked when setting it. S was already taken too D:

Tested the feature today through a movie with the mpv backend and worked great. Like I mentioned on the discussion post, I hadn't done any coding in Dart/Flutter before so here be dragons kinda.

## Issue Being Fixed
Fixes #497 

## Screenshots / Recordings
image

image

## Checklist

- [ ] For a future revision, it might be nice to have more flexibility on the file name.
- [ ] Testing on other OS's
- [ ] Make sure the feature doesn't show up on unsupported platforms.
- [ ] The MDK backend does image processing synchronously. In my testing it can cause a small stutter, so it'd probably be a good idea to move the processing somewhere it doesn't block (currently beyond my abilities)


lib/screens/settings/player_settings_page.dart
```diff
@@ -136,6 +140,101 @@ class _PlayerSettingsPageState extends ConsumerState<PlayerSettingsPage> {
...
+          trailing: SizedBox(
+              width: 100,
+              child: IntInputField(
+                controller: TextEditingController(text: ref.watch(videoPlayerSettingsProvider.notifier).screenshotNamePadding.toString()),
+                onSubmitted: (value) {
+                  ref.read(videoPlayerSettingsProvider.notifier).update(
+                    (current) => current.copyWith(screenshotNamePadding: value ?? 3),
+                  );
+
+                  ref.read(videoPlayerSettingsProvider.notifier).setscreenshotNamePadding(value ?? 3);
```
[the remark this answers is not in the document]
RainbowCookie32: Copy-pasted from somewhere else, removed the `update` one

lib/screens/settings/player_settings_page.dart
```diff
@@ -136,6 +140,97 @@ class _PlayerSettingsPageState extends ConsumerState<PlayerSettingsPage> {
...
+                              String? selectedDirectory = await FilePicker.platform.getDirectoryPath(
+                                  dialogTitle: context.localized.screenshotsPath, initialDirectory: screenshotsFolder);
+                              if (selectedDirectory != null) {
+                                ref.read(videoPlayerSettingsProvider.notifier).setScreenshotsPath(selectedDirectory);
+                              }
+                              Navigator.of(context).pop();
+                            },
+                            child: Text(context.localized.change),
+                          )
+                        ],
```
PartyDonut: This should be AdaptiveLayout.of(context).isDesktop if you only want to show it for desktop.

PartyDonut: I think it might also be a better idea to use the already existing download path in Settings->Fladder->Downloads path.

Currently file syncing syncs to that folder and uses "Synced" as a subfolder. We could use "Screenshots" as a subfolder for all screenshots.

lib/screens/video_player/video_player_controls.dart
```diff
@@ -679,6 +682,59 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
     }
   }
 
+  void takeScreenshot({required bool withSubtitles}) async {
```
PartyDonut: Let's move this function outside of the video player controls into it's own provider.

lib/wrappers/players/lib_mdk.dart
```diff
@@ -159,6 +161,41 @@ class LibMDK extends BasePlayer {
     return -1;
   }
 
+  @override
+  Future<Uint8List?> takeScreenshot(ScreenshotFormat format, bool withSubtitles) async {
```
PartyDonut: MDK seems to always take screenshots without subtitles?

Lets also change the track to disabled instead of loading/reloading the entire list. Currently the mdk player stutters quite heavily when taking a screenshot this might be part of the reason.

lib/wrappers/players/lib_mdk.dart
```diff
@@ -159,6 +161,41 @@ class LibMDK extends BasePlayer {
...
+    }
+
+    final snapshotData = await _controller?.snapshot();
+    final videoCodec = _controller?.getMediaInfo()?.video?[0].codec;
+
+    if (!withSubtitles && activeSubtitles != null) {
+      _controller?.setSubtitleTracks(activeSubtitles);
+    }
+
+    if (snapshotData != null && videoCodec != null) {
```
PartyDonut, line 169: Why not just return the Uint8List from the snapshot() call? 
Encoding the image here into a new set of bytes just seems like wasted cpu cycles.
Snapshot already returns the appropriate data including width, height and color channels (rgba).

pubspec.yaml
```diff
@@ -73,6 +73,8 @@ dependencies:
   fvp: ^0.33.1
   video_player: ^2.10.0
 
+  image: ^4.5.4 # For saving screenshots.
```
PartyDonut, line 75: We probably don't need this package dependency if we change the lib_mdk.dart functionality.

PartyDonut (changes requested):
Thanks for taking a shot at this. Functionality wise it seems to work just great!.
I've left a few comments let me know if any of them are unclear.

Some other notes/suggestions

Do we need the format selection? Would it not be better to always use PNG and let the user decide or convert it themselves later on. This would also clean up the user settings.

I would also suggest to use the existing download path directory and use "Screenshots" as sub directory. 

One other thing taking a screenshot gives the user no feedback, we should add some sort of message or pop-up that notifies the user that screenshot "001.png" was saved.

lib/providers/settings/video_player_settings_provider.dart
```diff
@@ -59,6 +62,8 @@ class VideoPlayerSettingsProviderNotifier extends StateNotifier<VideoPlayerSetti
     );
   }
 
+  VideoPlayerSettingsModel update(VideoPlayerSettingsModel Function(VideoPlayerSettingsModel currentState) value) => state = value(state);
```
PartyDonut: Please use a specific function call for this. Instead of changing and setting the entire model

lib/providers/settings/video_player_settings_provider.dart
```diff
@@ -85,6 +90,21 @@ class VideoPlayerSettingsProviderNotifier extends StateNotifier<VideoPlayerSetti
   void setBufferSize(int? value) => state = state.copyWith(bufferSize: value ?? 32);
   void setFitType(BoxFit? value) => state = state.copyWith(videoFit: value ?? BoxFit.contain);
 
+  String? get _screenshotsFolder => !kIsWeb && (Platform.isWindows || Platform.isLinux || Platform.isMacOS)
+      ? ref.read(videoPlayerSettingsProvider.select((value) => value.screenshotsPath))
+      : "";
+
+  String? get screenshotsFolder => _screenshotsFolder;
```
PartyDonut: These getters could have been one getter

lib/providers/settings/video_player_settings_provider.dart
```diff
@@ -85,6 +90,21 @@ class VideoPlayerSettingsProviderNotifier extends StateNotifier<VideoPlayerSetti
...
+  void setScreenshotsPath(String? path) => state = state.copyWith(screenshotsPath: path);
+
+  ScreenshotFormat get _screenshotFormat => ref.read(videoPlayerSettingsProvider.select((value) => value.screenshotFormat));
```
PartyDonut: These getters should all be moved to the VideoPlayerSettingsModel, also don't read the provider inside the provider itself.

RainbowCookie32:
Thanks for the review! I'll be chipping away at the comments later today, or tomorrow.

> Do we need the format selection? Would it not be better to always use PNG and let the user decide or convert it themselves later on. This would also clean up the user settings.

Personally I always use PNG. I only added the setting for user choice, and maybe allowing support for more formats in the future, but I'm definitely not attached to it.

> I would also suggest to use the existing download path directory and use "Screenshots" as sub directory.

Do you mean as a default, or just fixed to that folder? Personally I'd like to keep the option to select the directory where my screenshots go, mostly to keep them at hand and in the folder I had all the others.

> One other thing taking a screenshot gives the user no feedback, we should add some sort of message or pop-up that notifies the user that screenshot "001.png" was saved.

Fair enough, I'll take a look and try to put something together

lib/wrappers/players/lib_mdk.dart
RainbowCookie32: Well, mdk just returns an rgba array, it has to be encoded into a format (png/jpg in this case). It's nice for flexibility, but it forced me to bring in the image dependency to encode it. Didn't need it for mpv because it already returns png/jpg arrays, or optionally a bgra (ofc they are different) one.

lib/wrappers/players/lib_mdk.dart
RainbowCookie32: > MDK seems to always take screenshots without subtitles?

It seems to be inconsistent, befitting of a jank approach like this. In my tests, it usually takes the screenshot without subtitles, but it fails to enable them again. Sometimes it *does* enable them back though...

> Lets also change the track to disabled instead of loading/reloading the entire list. Currently the mdk player stutters quite heavily when taking a screenshot this might be part of the reason.

Hints welcome on how to approach this one

RainbowCookie32:
Thank god, I didn't break the PR after the rebase lol.

I pushed the first batch of changes and fixes. Still missing getting takeScreenshot a new home, plus the other things you mentioned that I replied to

PartyDonut:
> Thanks for the review! I'll be chipping away at the comments later today, or tomorrow.
> 
> > Do we need the format selection? Would it not be better to always use PNG and let the user decide or convert it themselves later on. This would also clean up the user settings.
> 
> Personally I always use PNG. I only added the setting for user choice, and maybe allowing support for more formats in the future, but I'm definitely not attached to it.

The only reason really is that this is a "nice to have" not so much a big requirement of a Jellyfin frontend.
So for that reason I would like to keep the amount of additional settings to a minimum, maybe we only need the screenshot hotkeys and keep everything else to a sane default?

For defaults we could use what you have now.
Format = PNG.
Name padding = 3?

Maybe it's a feature a lot of people will use don't know but those defaults will probably cater to most people.

> 
> > I would also suggest to use the existing download path directory and use "Screenshots" as sub directory.
> 
> Do you mean as a default, or just fixed to that folder? Personally I'd like to keep the option to select the directory where my screenshots go, mostly to keep them at hand and in the folder I had all the others.
> 

I would say fixed, so it always saves to "$downloadsPath/Screenshots"/, to prevent the application from having to handle multiple paths for saving different kinds of data.

All that being said it is not that I don't like the options I am just scared if we keep adding advanced settings like this we will and up with huge models and huge settings screens.

RainbowCookie32:
Sorry for disappearing! Lack of time/sleep/willpower took turns on controlling my brain this week.

Nuked the settings, replaced for the defaults and using the Downloads path for screenshots in its own folder. Got over keeping it as soon as I remembered symlinks are a thing :)

I noticed that the hotkey for taking screenshots without subtitles isn't being saved for some reason. It works fine after being changed, but it resets to the default after closing Fladder. Not sure what's going on there.

I'm aware I'm still dancing around *the other issues* - I intend to tackle them in the next few days (or try at least).

RainbowCookie32:
> I noticed that the hotkey for taking screenshots without subtitles isn't being saved for some reason. It works fine after being changed, but it resets to the default after closing Fladder. Not sure what's going on there.

Seemingly fixed itself somehow, won't complain

lib/models/settings/video_player_settings.dart
```diff
@@ -48,11 +50,25 @@ enum VideoHotKeys {
...
       VideoHotKeys.fullScreen => context.localized.fullScreen,
       VideoHotKeys.skipMediaSegment => context.localized.skipMediaSegment,
+      VideoHotKeys.takeScreenshot => context.localized.takeScreenshot,
+      VideoHotKeys.takeScreenshotClean => context.localized.takeScreenshotClean,
       VideoHotKeys.exit => context.localized.exit,
     };
   }
 }
 
+enum ScreenshotFormat {
```
PartyDonut: We can probably remove this now that it's not used

PartyDonut:
> Sorry for disappearing! Lack of time/sleep/willpower took turns on controlling my brain this week.
> 
> Nuked the settings, replaced for the defaults and using the Downloads path for screenshots in its own folder. Got over keeping it as soon as I remembered symlinks are a thing :)
> 
> I noticed that the hotkey for taking screenshots without subtitles isn't being saved for some reason. It works fine after being changed, but it resets to the default after closing Fladder. Not sure what's going on there.
> 
> I'm aware I'm still dancing around _the other issues_ - I intend to tackle them in the next few days (or try at least).

That is perfectly fine, no rush take your time 👍🏼.

lib/screens/video_player/video_player_controls.dart
RainbowCookie32: Would it be fine to put this on the video_player_provider?

RainbowCookie32:
I moved takeScreenshot out of video_player_controls to video_player_provider, which may or may not be good. It feels a bit more appropriate at least.
While I was there, I also made it return a bool to check for success for the _brand new screenshot indicator_ to display on screen. I based it off the seek indicator. I can't quite get the duration right, I feel like 500ms might be too fast but I tried higher numbers and it felt like it overstays its welcome. Might just be a matter of getting used to it

RainbowCookie32:
Forgot to add the new strings to *the big file*.

An issue I noticed since I added the screenshots indicator is that the seek key shortcut doesn't work anymore. If I take out the screenshot indicator, it works, and if I move the seek indicator to be after the screenshot one on `video_player_controls.dart`, then the latter one breaks. *Clearly* I broke something with my approach, but I can't tell what that something is (I'm also still failing to get mdk to play along with the clean screenshots, but I haven't dedicated *that* much time to it tbf).

PartyDonut:
> Forgot to add the new strings to _the big file_.
> 
> An issue I noticed since I added the screenshots indicator is that the seek key shortcut doesn't work anymore. If I take out the screenshot indicator, it works, and if I move the seek indicator to be after the screenshot one on `video_player_controls.dart`, then the latter one breaks. _Clearly_ I broke something with my approach, but I can't tell what that something is (I'm also still failing to get mdk to play along with the clean screenshots, but I haven't dedicated _that_ much time to it tbf).

The input handler can be a bit strange when nesting multiple listeners with autoFocus enabled. My first suggestion would be play around with who does/does not have autoFocus. But this might also require more investigating on my end to improve stability.

For disabling the subtitles in mdk I would suggest looking at the LibMDK class, specifically the setSubtitleTrack function. Easiest would be to use that function, save the current model, set the subtitle to (SubStreamModel.no()). And after the screenshot, set the original subtitle again. That should work.

RainbowCookie32:
> The input handler can be a bit strange when nesting multiple listeners with autoFocus enabled. My first suggestion would be play around with who does/does not have autoFocus. But this might also require more investigating on my end to improve stability.

I tried setting autoFocus to true on just one of them, on both, and then both on false and the result is still the same. Hotkey for screenshots works, but the seek ones don't.

> For disabling the subtitles in mdk I would suggest looking at the LibMDK class, specifically the setSubtitleTrack function. Easiest would be to use that function, save the current model, set the subtitle to (SubStreamModel.no()). And after the screenshot, set the original subtitle again. That should work.

I see `setSubtitleTrack` takes a `SubStreamModel` and `PlaybackModel`. Is there any way I can get the current models from LibMDK? For the sub one I thought about just saving it when the function is called, but idk if that'd work well for the playback one, since I assume it can also be modified from other places. My other idea is adding a `PlaybackModel` argument to takeScreenshot, but it'd be nice if it could be avoided imo as it's a generic method and only MDK would use it.

PartyDonut:
> I tried setting autoFocus to true on just one of them, on both, and then both on false and the result is still the same. Hotkey for screenshots works, but the seek ones don't.
> 

This is probably a tree ordering issue. The last one in the tree gets focus, however normally it should pass on any key that is not handled so that could be the problem. But looking at your code that seems to be handled just fine.

> I see `setSubtitleTrack` takes a `SubStreamModel` and `PlaybackModel`. Is there any way I can get the current models from LibMDK? For the sub one I thought about just saving it when the function is called, but idk if that'd work well for the playback one, since I assume it can also be modified from other places. My other idea is adding a `PlaybackModel` argument to takeScreenshot, but it'd be nice if it could be avoided imo as it's a generic method and only MDK would use it.

The players should not be aware of the selected subtitles or at least not as they are presented in the UI. My suggestion would be to set the subtitles just before you ask the player to take the screenshot and then implement it like in the video_player_options_sheet.dart -> showSubSelection() function. 

Keep in mind disabling subtitles while transcoding could cause the playback the re-evaluate whether transcoding is needed. Make sure you keep the subtitles state as is when transcoding.

RainbowCookie32:
I feel like it's a miracle I haven't borked this PR yet with the jank rebases.

I think I got the clean screenshots working on both players now, did a quick test and they are looking fine. Subs noticeably blink for a second now on mpv, which didn't happen before and is an unfortunate nerf, but I think this is the cleanest way to get feature parity with both players.

The seek hotkeys still refuse to play ball alongside with the screenshot ones tho...

PartyDonut:
Just checking in @RainbowCookie32 are you still interested in finishing this?

RainbowCookie32:
Oof, it's been a while. Seems like I have a few rebases pending.

I think that the last batch of changes I pushed addressed everything you mentioned, with only the hotkeys conflict (seek and screenshots) being left, which I never really managed to figure out. Do let me know if there's anything else I should look into tho, I'll take a look at it

RainbowCookie32:
Re-rebased, didn't realise I could request a new review so there it goes!

RainbowCookie32:
Nice! Thanks for the guidance on this too :)

PR #556, opened 2025-11-01
fix: Bitrate display in Media Info

matt-hb:
## Pull Request Description

Tiny fix for the bitrate info in the Movie/Episode Info window, and also I have formatted both audio and video to be what I would expect, (for high bitrate video, over 10 Mbps it is `xx.x Mbps`, under that it is the value in kbps, and for any audio, the bitrate in kbps), open to opinions, might even be something the user could set for themselves via a toggle.

## Issue Being Fixed

The Jellyfin API returns the video and audio bitrate in `bps`, but previously they were displayed as though they are `kbps` values.
(+ formatting)

## Screenshots / Recordings

High bitrate ex:
image

Lower bitrate ex:
image

## Checklist

### DIFF
--- a/lib/util/bitrate_formatting.dart
+++ b/lib/util/bitrate_formatting.dart
@@ -8,16 +8,16 @@ extension BitrateFormats on int? {
   }
 
   String? get videoBitrateFormat {
-    const int VIDEO_HIGH_BITRATE_CUTOFF = 10000000;
-    const int Kb = 1000;
-    const int Mb = Kb * Kb;
+    const int highBitrateCutoff = 10000000;
+    const int kb = 1000;
+    const int Mb = kb * kb;
 
     final bitrate = this;
     if (bitrate == null) return null;
-    if (bitrate >= VIDEO_HIGH_BITRATE_CUTOFF) {
+    if (bitrate >= highBitrateCutoff) {
       return "${(bitrate / Mb).toStringAsFixed(1)} Mbps";
     } else {
-      return "${(bitrate / Kb).round()} kbps";
+      return "${(bitrate / kb).round()} kbps";
     }
   }
 }


lib/models/information_model.dart
```diff
@@ -38,7 +38,9 @@ class InformationModel {
               "Aspect Ration": e.aspectRatio,
               "Interlaced": e.isInterlaced,
               "FrameRate": e.realFrameRate,
-              "Bitrate": "${e.bitRate} kbps",
+              "Bitrate": e.bitRate! <= 10000000
```
PartyDonut: The bang operator is something best avoided.

While the logic itself is perfectly fine. It would be better to separate this into it's own extension function. That way we can also keep it null safe.

Take a look at the "size_formatting.dart" file for where and how to implement a very similar extension function.

matt-hb: Good call, thanks for the heads up.

lib/util/bitrate_formatting.dart
```diff
@@ -0,0 +1,23 @@
...
+
+extension BitrateFormats on int? {
+  String? get audioBitrateFormat {
+    final bitrate = this;
+    if (bitrate == null) return null;
+    return "${(bitrate / 1000).round()} kbps";
+  }
+
+  String? get videoBitrateFormat {
+    const int VIDEO_HIGH_BITRATE_CUTOFF = 10000000;
```
PartyDonut: Seems like the AI suggested the naming convention of this, not sure why. But in dart for local constants its normal to use camelCase.

As I'm already nitpicking should the Kb variable in this case not also be lowercase 'kb' same as the returned string?

matt-hb: haha, the AI suggestion + my background made me go against the linter suggesting camelCase for consts, I'll change that over.
Very good nitpicking eye on the 'kb' name as well, just made both Mb and kb the same on muscle memory 👍

Issue #565, opened 2025-11-02
🐛 Playback not continuing from where I left off Android Mobile


schembriaiden:
### Description

As from 0.8.0 playback on Android does not continue from where I left off  when the app is first opened. Instead it starts playback from the beginning

### Reproduction steps

```bash
1. Close the app fully
2. Reopen the app
3. Try to continue playback from where you left off before
4. It will proceed to start from the beginning
```

### Logs

```bash

```

### Platform

Android

### App Version

0.8.0

schembriaiden:
This has been resolved with the following commit: https://github.com/DonutWare/Fladder/commit/2f8d61bb6adb488f9602082ba267a80502939fac

schembriaiden:
Reopening as it is still happening but not as frequent

PR #571, opened 2025-11-02
fix: Playback continues from previous time

schembriaiden:
## Pull Request Description

This pull request updates the logic for starting video playback in the `VideoPlayerNotifier` class to improve the handling of audio and subtitle track selection and seeking behavior. The main change is that audio and subtitle tracks are only reset when playback starts from the beginning, and seeking now waits for playback to actually start.

Playback initialization improvements:

* Only reset audio and subtitle tracks (`setAudioTrack` and `setSubtitleTrack`) if playback starts from the beginning (`Duration.zero`), preventing unnecessary track changes when resuming from a specific position.
* Playback is started before seeking to a non-zero start position, and a delay is added to ensure playback has begun before seeking, improving reliability of resume behavior.

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/issues/565

## Checklist


lib/providers/video_player_provider.dart
```diff
@@ -129,12 +130,53 @@ class VideoPlayerNotifier extends StateNotifier<MediaControlsWrapper> {
         null,
         onDone: () async {
           final start = startPosition;
+          
+          if (start == Duration.zero) {
```
PartyDonut, line 134: Why only when starting from the beginning? loadPlaybackItem is only called when new media is loaded so it should always load the subtitle/audio tracks.

lib/providers/video_player_provider.dart
```diff
@@ -129,12 +130,53 @@ class VideoPlayerNotifier extends StateNotifier<MediaControlsWrapper> {
...
+            await state.setSubtitleTrack(null, model);
+          }
+          
+          await state.play();
+          
           if (start != Duration.zero) {
-            await state.seek(start);
+
+            final playStartTime = DateTime.now();
+            const minDelay = Duration(milliseconds: 800);
```
PartyDonut, line 145: This seems like a very elaborate fix for what could probably be solved by doing this.

 if (start != Duration.zero) {
    await Future.delayed(const Duration(milliseconds: 500));
    await state.seek(start);
  }
  await state.play();

PartyDonut (changes requested):
This seems more like a workaround then a fix.

Seems like it only happens with the mpv player and only on the first video load. Would probably be better to see why the mpv player has problems with this all of a sudden.

schembriaiden:
I will look into this in the coming days to try and find the actual cause of the issue, been a bit busy this week.

PartyDonut:
@vikingnope

Maybe something to have a look at I have updated the media_kit dependencies. And one of the commits mentioned seek not properly seeking to the media's new start position.

Given that the problem occurred without any changes it's worth checking out if this is fixed with the new update.

https://github.com/DonutWare/media-kit/commit/d310049f24196250d876efb02b9ff56fa9ef5068

schembriaiden:
This issue seems to be resolved with the latest changes you made with media-kit. Thanks for checking this out and notifying me. :)

Issue #451, opened 2025-08-13
🐛 Download progress stuck at 0%, then completes


stignarnia:
### Description

Basically the title, of course I can provide further information

### Reproduction steps

```bash
Sync any media
See the progress bar in the sync page
Monitor networking speed to see that actually it is downloading successfully
After some time the media will appear as synced
```

### Logs

```bash

```

### Platform

Android

### App Version

0.7.5 and 0.7.0

PartyDonut:
Hi, if you could check the logs in Setings -> About and report back with any errors related to downloads that would help 👍🏼

stignarnia:
Hi! I cleared the logs and then downloaded something, this is now the only entry:

```
[SEVERE  |  Flutter error: Null check operator used on a null value, 2025-08-15T16:35:00.829776

#0      State.context (package:flutter/src/widgets/framework.dart:959)
#1      _PosterImageState.pressedWidget (package:fladder/screens/shared/media/components/poster_image.dart:76)
<asynchronous suspension>
]
```

PartyDonut:
That log is not related to downloads/syncing unfortunately. Do both the in-app progress and the notification itself show no progress?

stignarnia:
For sure inside the app, I will check the notification because I kept them off so far

stignarnia:
![Image](https://github.com/user-attachments/assets/e85c5b9a-be5d-4564-9a5e-2fc1d8f468b3)

The notification is stuck like this and then completes. As you can see in the top right, internet is very much in use. Also it's not trying to calculate the size of the download like Chrome sometimes does because the size is already shown in the Sync page.

PartyDonut:
Strange that  it's stuck at 50% in the notification. Not sure where to start with this, if there is any more info you could provide related to the phone/os/installation/vpn you are using that would help.

stignarnia:
No no it's not stuck at 50, it's doing that thing where the progress bar moves around, similar to when chrome says "calculating". In fact as you can see the download speed is not populated, it says --.
My tablet is android 15, both pixel software on my phone and OnePlus, the installation of jellyfin is the barebones installer for Ubuntu but installed in an lxc container on proxmox with gpu passthrough, and to connect I just a use an authentik reverse proxy with authentication disabled (I have other apps for which I need authentication)

benjamin-stern:
[...]

PR #583, opened 2025-11-06
fix: Sync basename parsing for different hosts

### LINKED ISSUE #457: 🐛 Sync doesn't appear to work
### Description

Trying to sync videos. The toast message appears but disappears. Looking in synced items shows the files but states "Not found". Nothing is appearing in logs at all.

### Reproduction steps

```bash
1. Go to any library item
2. Click on sync
```

### Screenshots

_No response_

### Logs

```bash
Nothing appears in logs
```

### Platform

Android

### App Version

v0.7.5

benjamin-stern:
## Pull Request Description

I tested the latest release of 8.0 and saw that this issue still hasn't been resolved, so I thought I would take a whirl at it.

Fixes the issue with malformed filename where it seems there are slashes in the file's name, this sanitizes it if for some reason it has occurred.

It updates and persists the name and includes it in the download task.

I've tested and confirmed the fix on the windows development build (unfortunately I don't have the environment setup to test android)

## Issue Being Fixed

This fixes the issues where the files cannot be synced for some users.

Resolves #457 #451

## Checklist

### DIFF
--- a/lib/providers/sync_provider.dart
+++ b/lib/providers/sync_provider.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'dart:developer';
 import 'dart:io';
 
+import 'package:fladder/util/string_extensions.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart' hide ConnectionState;
 
@@ -503,19 +504,11 @@ class SyncNotifier extends StateNotifier<SyncSettingsModel> {
         await ref.read(backgroundDownloaderProvider).cancelTaskWithId(currentTask.id);
       }
       if (!skipDownload) {
-        // Sanitize filename to avoid path separators causing ArgumentError.
-        final rawFilename = syncItem.videoFileName;
-        final safeFilename = path.basename(rawFilename ?? 'unknown').replaceAll(RegExp(r'[\\/]+'), '_');
-        if (safeFilename != rawFilename) {
-          // Update syncItem expectation so completion detection (videoFile.existsSync) matches saved file.
-          syncItem = syncItem.copyWith(videoFileName: safeFilename);
-          await updateItem(syncItem); // persist new filename
-        }
         final downloadTask = DownloadTask(
           taskId: syncItem.id,
           url: Uri.parse(downloadUrl).toString(),
           directory: syncItem.directory.path,
-          filename: safeFilename,
+          filename: syncItem.videoFileName,
           updates: Updates.statusAndProgress,
           baseDirectory: BaseDirectory.root,
           urlQueryParameters: {"api_key": user.credentials.token},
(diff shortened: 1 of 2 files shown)


aero31aero:
The code looks good, but I have a question. In the issue, the file name I was failing to sync didn't have forward slashes in it.. how does it change the situation there?

(Also, I'm a noob at Flutter, is there an easy way for me to test this PR locally and report back?)

benjamin-stern:
@aero31aero I'm not sure what systems you are running on but I suspect if you have the issue on one client you'll probably have it on all clients. For me personally I have the issue with Android and Windows so I developed and tested on my windows machine.

[Windows Build](https://drive.google.com/file/d/1xIEI5JuO6xBcRPWSS8eJvklWAr8qjdEs/view?usp=sharing)

You can click on the above link to download the windows build and test that it functions correctly. (works for me)

I suspect it's cause by my Jellyfin Server being installed on a windows machine and so it retrieves the path as part of the file location and tries to nest it internally, instead i just tell it to get the filename and strip out any `\` characters if they remain and make that the file and path local when downloading and saving.

lib/providers/sync_provider.dart
```diff
@@ -503,11 +503,19 @@ class SyncNotifier extends StateNotifier<SyncSettingsModel> {
         await ref.read(backgroundDownloaderProvider).cancelTaskWithId(currentTask.id);
       }
       if (!skipDownload) {
+        // Sanitize filename to avoid path separators causing ArgumentError.
+        final rawFilename = syncItem.videoFileName;
+        final safeFilename = path.basename(rawFilename ?? 'unknown').replaceAll(RegExp(r'[\\/]+'), '_');
+        if (safeFilename != rawFilename) {
+          // Update syncItem expectation so completion detection (videoFile.existsSync) matches saved file.
+          syncItem = syncItem.copyWith(videoFileName: safeFilename);
+          await updateItem(syncItem); // persist new filename
```
PartyDonut: Instead of sanitizing the name here and only when downloading. It would be better to always sanitize it inside of the "createSyncItem" function. That way we don't loose the file reference for any reason.

Also please don't use comments when it's clear what is happening. Create a new extension function in string_extensions.dart and give it a suitable name.

PartyDonut (changes requested):

PartyDonut:
First of thanks for implementing this 👍🏼 .

While sanitizing the file is a good thing, like @aero31aero mentioned I'm also not quite sure this fixes the issues some people where having. 

So let's merge it after the requested changes. But will probably have to check in with the issue creator whether this fixes it for them.

benjamin-stern:
@PartyDonut right part of the benefit is that I was also experiencing the bug on all my clients both windows/android (multiple clients), although I only built and tested for windows i expect the fix to affect all clients positively.

As requested I moved the functionality to the string extensions and I've found that it works like the path.basename but support non-linux type servers and since I was running jellyfin on a windows server now the download function appears to work as expected.

It would be great to hear feedback from others that it's working as expected (including for iOS platforms) I will test the android when it's merged.

Thank you for your attention to the matter in trying to help resolve.
🙏

lib/providers/sync_provider.dart
benjamin-stern: Done

PR #607, opened 2025-11-12
fix: use /Items instead of /Latest with includeItemTypes parameter to massively improve latency

### LINKED ISSUE #590: 🐛 Navigating to the "Libraries" page makes Jellyfin 10.11 unresponsive for several minutes.
### Description

I noticed that both Streamyfin and Fladder made my server unusable for several minutes under certain conditions, and I tracked it down to the `/Items/Latest?userId=XXX&parentId=XXX&includeItemTypes=Series&isPlayed=false&imageTypeLimit=1&limit=14` API request in the Libraries page. After some research, I found https://github.com/jellyfin/jellyfin/issues/15014#issuecomment-3443953613 which suggests that the `includeItemTypes` parameter for whatever reason makes Jellyfin completely unresponsive for several minutes as of 10.11 and has not yet been fixed.

Fladder is great and I really want to use it, but for now the Libraries page is a self-destruct button due to Jellyfin's bug, so I ask that as a workaround, something like https://github.com/streamyfin/streamyfin/pull/1141 is implemented.

This bug is present on all three versions of Fladder I tried (web, android, android tv).

### Reproduction steps

```bash
1. Update to Jellyfin 10.11.0-2
2. Click on Libraries
3. Watch the server become unresponsive
```

### Screenshots

_No response_

### Logs

```bash

```

### Platform

Web

### App Version

0.8.0

gabehf:
## Pull Request Description

This PR changes the `/Items/Latest` endpoint to the `/User/XXX/Items` endpoint in order to avoid the massive slowdown that occurs when using `includeItemTypes` on the `/Latest` endpoint with some libraries. In my testing, this change is ~100x faster (1m20s vs 800ms).

## Issue Being Fixed

The libraries page calls `/Items/Latest?includeItemTypes=...` which, for some libraries, will cause the server to become unresponsive for several minutes with large TV libraries due to some bug in Jellyfin that has not yet been fixed. Changing the API request to `/Items?includeItemTypes` and sorting by date content added, the issue can be avoided.

Resolves #590

## Screenshots / Recordings

N/A

## Checklist

### DIFF
--- a/lib/providers/library_screen_provider.dart
+++ b/lib/providers/library_screen_provider.dart
@@ -165,6 +165,7 @@ class LibraryScreen extends _$LibraryScreen {
       parentId: viewModel.id,
       sortBy: [ItemSortBy.datelastcontentadded, ItemSortBy.datecreated, ItemSortBy.sortname],
       sortOrder: [SortOrder.descending],
+      limit: 9,
       includeItemTypes: viewModel.collectionType.itemKinds.map((e) => e.dtoKind).toList(),
     );
     newRecommendations = [
--- a/lib/providers/service_provider.dart
+++ b/lib/providers/service_provider.dart
@@ -700,6 +700,7 @@ class JellyService {
     String? parentId,
     List<ItemSortBy>? sortBy,
     List<SortOrder>? sortOrder,
+    int? limit,
     bool? recursive,
     List<BaseItemKind>? includeItemTypes,
   }) async {
@@ -710,6 +711,7 @@ class JellyService {
       sortBy: sortBy,
       sortOrder: sortOrder,
       includeItemTypes: includeItemTypes,
+      limit: limit,
     );
   }
 


lib/providers/library_screen_provider.dart
```diff
@@ -161,17 +161,17 @@ class LibraryScreen extends _$LibraryScreen {
       ];
     }
 
-    final latest = await api.usersUserIdItemsLatestGet(
+    final latest = await api.usersUserIdItemsGet(
       parentId: viewModel.id,
-      limit: 9,
-      imageTypeLimit: 1,
+      sortBy: [ItemSortBy.datelastcontentadded, ItemSortBy.datecreated, ItemSortBy.sortname],
```
PartyDonut, line 166: Is there a reason why the limit is removed?

PartyDonut (changes requested):
Looks good to me. Small change request/question.

I'm also going to assume that including the userId in the path instead of queryParam did not help for the "_usersUserIdItemsLatestGet" api call?

lib/providers/library_screen_provider.dart
gabehf: The usersUserIdItemsGet function (method? i don't know dart/flutter) did not have a limit parameter, so I didn't include it because I'm not super well versed in the Jellyfin API. My testing is without the limit parameter and it is still very significantly faster, but I can add it back if you prefer.

gabehf:
> I'm also going to assume that including the userId in the path instead of queryParam did not help for the "_usersUserIdItemsLatestGet" api call?

Truthfully I don't remember all of the different combinations of path/parameter I tested, only that this version of the request was the first one I found that did not cause the issue, so I ran with it.

lib/providers/library_screen_provider.dart
PartyDonut: "usersUserIdItemsGet" has a limit variable. It was just not exposed in the service_provider.dart. So lets keep the limit set to 9.

PartyDonut:
> > I'm also going to assume that including the userId in the path instead of queryParam did not help for the "_usersUserIdItemsLatestGet" api call?
> 
> Truthfully I don't remember all of the different combinations of path/parameter I tested, only that this version of the request was the first one I found that did not cause the issue, so I ran with it.

Let's test if the latest also works, and what exactly causes the "fix" that way we don't make changes we don't want. Because I'm not quite sure if/what difference there is between the "latest" api call and the "itemsGet" api call with these parameters.

gabehf:
I do know that the `/Latest` endpoint did not work with either the user ID in parameters or in the path, if that's what you are asking. The `/Items` endpoint works with the user ID in the path, but I did not test if it works with user ID in the parameters.

PartyDonut:
> I do know that the `/Latest` endpoint did not work with either the user ID in parameters or in the path, if that's what you are asking. 

Yes that was the question 👍🏼 . For now it seems fine functionally it should work the same. Only thing left is setting the limit to 9 so we don't fetch too much.

PartyDonut (approved):
LGTM. Thanks for checking out what fixes work.

PR #630, opened 2025-11-24
feat: Nix support for reproducible builds and dev shell

schembriaiden:
## Pull Request Description

This pull request adds comprehensive Nix support for the project, introducing a Nix flake for reproducible builds, a development shell, and automation for updating dependency hashes. The changes enable easier development and packaging on Nix-based systems, automate hash management for dependencies, and improve the developer onboarding experience.

**Nix Flake and Build Infrastructure:**

* Added a new `flake.nix` file that defines the project as a Nix flake, specifying package builds, dependency pinning, and custom source builders for `media_kit` and related plugins. This allows reproducible and declarative builds of the Flutter application on NixOS and compatible systems.
* Introduced custom Nix derivations for the `fvp` plugin (`nix/fvp.nix`) and `media_kit_libs_linux` (`nix/media_kit_libs_linux_donutware.nix`), ensuring all required native dependencies are fetched and built in a fully reproducible manner.

**Development Environment:**

* Added a `shell.nix` file that sets up a complete development environment with all necessary build tools, libraries, and environment variables for Flutter and Linux desktop development. This streamlines onboarding and ensures consistency across developer machines.
* Updated `.envrc` to automatically use the flake-based environment if Nix is installed, improving developer experience for those using `direnv`.

**Automation and CI:**

* Added a GitHub Actions workflow (`.github/workflows/update-nix-hashes.yml`) that automatically updates Nix hashes for `media_kit` when `pubspec.lock` changes, and creates a pull request with the updated hashes. This reduces manual maintenance and keeps builds reproducible.

## Issue Being Fixed

N/A

## Checklist

- [ ] If a new package was added, did you ensure it works for all supported platforms? Is the package well maintained
- [ ] Check that any changes are related to the issue at hand.

PR #631, opened 2025-11-25
docs: add install instructions for Nix, Homebrew, and Docker

schembriaiden:
## Pull Request Description

Added install instruction for:

- Nix
- Homebrew
- Docker Rootless

## Issue Being Fixed

N/A

Resolves #issue-number

## Checklist


PartyDonut:
Hi thanks for updating the documentation.

Just saw your comment on the license agreement for winget.

The reason I disabled it for now is I'm not sure whether the license itself is something I agree with. Looks to me like Microsoft is granted pretty much a do whatever they want with the source code license. So I'm not quite sure if I would still like to support winget.

schembriaiden:
> The reason I disabled it for now is I'm not sure whether the license itself is something I agree with. Looks to me like Microsoft is granted pretty much a do whatever they want with the source code license. So I'm not quite sure if I would still like to support winget.

Yep I totally agree with it. Honestly I don't really like the way winget does things especially security wise as I think it is the most unsecure package manager I have seen yet. I have removed tthe winget installation part and I added the Nix part, so if everythiing is fine this should be merged after https://github.com/DonutWare/Fladder/pull/630

PartyDonut:
Same as #630

PR #680, opened 2026-01-03
feat: Add tizen support

popdollar:
## Pull Request Description

Added support Tizen OS 6.0+ for Samsung TVs.  It is in a working state (at least on my TV S90C and the emulator).
It is build with Samsungs Flutter SDK.  
I've used tizens videohole video_player since they recommend using it when specifically targeting TVs. 

## Disclaimer

I've targeted main functionality and there might be changes that are breaking, since this have not been tested with any other OS. I don't believe the additional packages specifically for Tizen are breaking for any other clients, but I can not guarantee anything.

## Checklist

### DIFF
--- a/lib/screens/video_player/tizen_video_player_controls.dart
+++ b/lib/screens/video_player/tizen_video_player_controls.dart
@@ -189,7 +189,7 @@ class _TVControlsState extends ConsumerState<TVControls> {
         break;
       case 'seekBack':
         final backwardSpeed = 
-          ref.read(userProvider.select((value) => value?.userSettings?.skipBackDuration.inSeconds ?? 10));
+          ref.read(userProvider.select((value) => value?.userSettings?.skipBackDuration.inSeconds ?? 30));
         seekBack(ref, seconds: backwardSpeed);
         break;
       case 'playPause':
@@ -698,8 +698,8 @@ class _TVControlsState extends ConsumerState<TVControls> {
   }
 
   Widget seekBackwardButton(WidgetRef ref) {
-    final backwardSpeed = 10;
-        //ref.read(userProvider.select((value) => value?.userSettings?.skipBackDuration.inSeconds ?? 30));
+    final backwardSpeed = 
+      ref.read(userProvider.select((value) => value?.userSettings?.skipBackDuration.inSeconds ?? 30));
      return _wrapNav(
        'seekBack',
        IconButton(
@@ -901,36 +901,15 @@ class _TVControlsState extends ConsumerState<TVControls> {
         resetTimer();
         return true;
       }
-      // if (key == LogicalKeyboardKey.arrowUp) {
-      //   // Move up a few steps (best-effort grid movement)
-      //   _moveSelection(-3);
-      //   return true;
-      // }
-      // if (key == LogicalKeyboardKey.arrowDown) {
-      //   _moveSelection(3);
-      //   return true;
-      // }
-      // Activation keys
       
       if (key == LogicalKeyboardKey.select || key == LogicalKeyboardKey.enter || key == LogicalKeyboardKey.space) {
-        //activateSelection();
-        //_setActiveIndicator('Activate');
         resetTimer();
         return true;
       }
     }
 
-    // // Arrow keys -> seek indicator (when overlay not visible)
-    // if (key == LogicalKeyboardKey.arrowLeft || key == LogicalKeyboardKey.arrowRight) {
-    //   _setActiveIndicator('Seek');
-    //   // do not swallow here; other handlers (seek indicator) will process the key
-    //   return false;
-    // }
-
     // Up/Down -> volume (when overlay not visible)
     if (key == LogicalKeyboardKey.arrowUp || key == LogicalKeyboardKey.arrowDown || key == LogicalKeyboardKey.select || key == LogicalKeyboardKey.enter || key == LogicalKeyboardKey.space) {
-      // _setActiveIndicator('Volume');
-      // do not swallow here; allow normal processing
       toggleOverlay(value: true);
       return false;
     }
--- a/lib/wrappers/players/tizen_player.dart
+++ b/lib/wrappers/players/tizen_player.dart
@@ -82,9 +82,9 @@ class TizenPlayer extends BasePlayer {
   }
 
   @override
-  Future<void> pause() => _controller?.pause() ?? Future.value();
+  Future<void> pause() async => _controller?.pause();
   @override
-  Future<void> play() => _controller?.play() ?? Future.value();
+  Future<void> play() async => _controller?.play();
   @override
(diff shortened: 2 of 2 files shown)


lib/models/syncing/database_item.dart
```diff
@@ -205,4 +216,15 @@ class AppDatabase extends _$AppDatabase {
       ),
     );
   }
+
+  static LazyDatabase _openConnectionTizen() {
```
PartyDonut, line 220: What is the reason we need to open the database this way for tizen?

Are downloads even supported for on samsung tv's?

lib/util/resolution_checker.dart
PartyDonut, line 1: We can probably leave this out completely, or do we need to check for resolution changes on Samsung TV's?

lib/screens/video_player/tizen_video_player_controls.dart
PartyDonut, line 1: This file could use some clean-up. I'm also not a big fan of the way navigation is handled, we seem to be hard-coding a navigation path. 

It's fine to leave it as is for now if it works. But will probably update this with a more rigid like a FocusTraversalGroup solution. 
The native Flutter controls have to be updated for dPad navigation at some point any way.

Do remove the commented out code if it's no longer need though.

lib/wrappers/players/tizen_player.dart
```diff
@@ -0,0 +1,344 @@
...
+
+  @override
+  Future<void> open(BuildContext context) async {
+    await Navigator.of(context, rootNavigator: true).push(
+      MaterialPageRoute(builder: (_) => const video_screen.VideoPlayer()),
+    );
+  }
+
+  @override
+  Future<void> pause() => _controller?.pause() ?? Future.value();
```
PartyDonut:
```suggestion
  Future<void> pause() async => _controller?.pause();
```

Let's do this instead of returning a empty Future.value();

lib/wrappers/players/tizen_player.dart
```diff
@@ -0,0 +1,344 @@
...
+  @override
+  Future<void> play() => _controller?.play() ?? Future.value();
```
PartyDonut:
```suggestion
  Future<void> play() async => _controller?.play();
```

Same here.

lib/wrappers/players/tizen_player.dart
```diff
@@ -0,0 +1,344 @@
...
+  @override
+  Future<void> setVolume(double volume) async => _controller?.setVolume(volume / 100);
+  @override
+  Future<void> loop(bool loop) async => _controller?.setLooping(loop);
+
+  @override
+  Future<int> setAudioTrack(AudioStreamModel? model, PlaybackModel playbackModel) async {
+    final wantedAudioStream = model ?? playbackModel.defaultAudioStream;
+    if (wantedAudioStream == AudioStreamModel.no() || wantedAudioStream == null) {
+      return -1;
```
PartyDonut, line 103: This no longer allows for disabling audio. Is that something not supported by the Tizen player?

lib/wrappers/players/tizen_player.dart
PartyDonut, line 1: This file is probably the biggest "pain" point for merging this into the main repository. It also seems to be missing some imports but that could be because I'm using the default flutter sdk.

If we want to keep everything in a single repo my suggestion would be to rename this file to tizen_player.dart.orig. And create a new "stub" file named tizen_player.dart that extends the BasePlayer so we don't get any errors.

That way when working on Tizen we can simply swap the file out for the original and continue from there.

lib/main.dart
```diff
@@ -297,12 +307,32 @@ class _MainState extends ConsumerState<Main> with WindowListener, WidgetsBinding
     final scrollBehaviour = const MaterialScrollBehavior();
     return DynamicColorBuilder(
       builder: (ColorScheme? lightDynamic, ColorScheme? darkDynamic) {
-        final lightTheme = themeColor == null
+        final rawLightTheme = themeColor == null
```
PartyDonut, line 310: Do we need this change?

pubspec.yaml
PartyDonut, line 1: Probably best to create a new "tizen" only pubspec that we can then swap between when we need to debug/build tizen.

lib/main.dart
```diff
@@ -63,6 +67,12 @@ Future<Map<String, dynamic>> loadConfig() async {
 }
 
 void main(List<String> args) async {
+  if (isTizen) {
```
PartyDonut, line 70: Logic like this should be moved to a new file. So we can create a "stub" file like mentioned previously for the other platforms that implements the functions/boolean with nothing attached.

That also includes the following imports.

`import 'dart:ffi' as ffi; `
`import 'package:flutter_tizen/flutter_tizen.dart';`
`import 'package:sqlite3/open.dart' as sqlite3_open;`

While only the ffi is not supported on some platforms best to not use it at all by default if possible.

PartyDonut (changes requested):
Sorry for the late reply. And thank you for having a crack at getting Fladder to Samsung TV's.

I've left a few comments. Let me know if any of them are unclear.

The basic idea is we need to separate some of the calls that are needed for Tizen in new files. And create stub files for the other platforms to call.

There are probably some things I missed but let's fix the current comments and we can see where we go from there.

lib/models/syncing/database_item.dart
popdollar: Drift is not really supported with the samsung flutter SDK, so this is to keep the opening synchronous. Could be it's the wrong way to do it. This seems to be working with downloads. It has not been extensively tested. I've downloaded and played part of movie with it, that is about the extent of the testing.

lib/util/resolution_checker.dart
popdollar: No you're right, and it doesnt really work as intended anyways.

lib/screens/video_player/tizen_video_player_controls.dart
popdollar: I agree. It ended up being a bit of a mess with the code I wrote. I got it to a "working" state and sort of just forgot about it. It is in a usable state, but for sure not in a finished version. I am not sure it makes sense to have separate TV/Tizen controls if dpad navigation is added to the player. 

I'll remove the commented out code - my bad.

lib/wrappers/players/tizen_player.dart
popdollar: I could not get it to work, so I left it out. From what I can tell you would have to add an empty audio track and then use that. An ugly easy "fix" would be to just mute audio.

lib/main.dart
popdollar: Possibly not, but the theme caused a lot of errors because of unsupported animations. Very likely that it could be fixed in a better, less messy, way

Orbernator:
Just adding my two cents in here. Will this be released on to the Tizen App Store or will it have to be self built?

PR #711, opened 2026-01-14
feat: Display specials

HachiHalation:
## Pull Request Description

Adds a special features row to show details and season details screens, similar to the official jellyfin client. Unfortunately, due to how the special features API is set up, it is really difficult to group them by season in the show screen without making a bunch of API calls.

- Add SpecialFeatureModel
- Add SpecialFeatureRow widget (based on EpisodePosters)
- Add the required data to relevant providers
- Add Special Feature localization string to EN file

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/discussions/632

## Screenshots / Recordings

show screen:

image

season screen:

image

## Checklist

### DIFF
--- a/lib/screens/details_screens/season_detail_screen.dart
+++ b/lib/screens/details_screens/season_detail_screen.dart
@@ -1,6 +1,5 @@
 import 'package:fladder/screens/shared/media/special_features_row.dart';
 import 'package:fladder/util/item_base_model/item_base_model_extensions.dart';
-import 'package:fladder/util/item_base_model/play_item_helpers.dart';
 import 'package:flutter/material.dart';
 
 import 'package:flutter_riverpod/flutter_riverpod.dart';


lib/models/items/special_feature_model.dart
```diff
@@ -0,0 +1,121 @@
...
+import 'package:fladder/models/items/series_model.dart';
+import 'package:fladder/util/localization_helper.dart';
+import 'package:fladder/util/string_extensions.dart';
+
+part 'special_feature_model.mapper.dart';
+
+@MappableClass()
+class SpecialFeatureModel extends ItemStreamModel with SpecialFeatureModelMappable {
+  final List<Chapter> chapters;
+  final ItemLocation? location;
```
PartyDonut: This seems unused probably better to remove it.

lib/models/items/special_feature_model.dart
```diff
@@ -0,0 +1,121 @@
...
+        canDelete: item.canDelete,
+        canDownload: item.canDownload,
+        mediaStreams: MediaStreamsModel.fromMediaStreamsList(item.mediaSources, ref),
+        jellyType: item.type,
+      );
+
+  static List<SpecialFeatureModel> specialFeaturesFromDto(List<dto.BaseItemDto>? dto, Ref ref) {
+    return dto?.map((e) => SpecialFeatureModel.fromBaseDto(e, ref)).toList() ?? [];
+  }
+}
```
PartyDonut: Please add the newline as well. So we can keep consistent formatting.

lib/providers/service_provider.dart
```diff
@@ -675,6 +675,16 @@ class JellyService {
     return response.body?.items?.map((e) => ItemBaseModel.fromBaseDto(e, ref)).toList() ?? [];
   }
 
+  Future<Response<List<BaseItemDto>>> itemsItemIdSpecialFeaturesGet({
+    required String itemId
+  }) async {
+    try {
+      return api.itemsItemIdSpecialFeaturesGet(itemId: itemId, userId: account?.id);
+    } catch (e) {
+      return Response(http.Response("", 400), []);
```
PartyDonut: Any errors here will be obfuscated because we throw away whatever the exception is telling us.
Remove the try{} catch{}. So we can monitor the original response.

lib/screens/shared/media/special_features_row.dart
```diff
@@ -0,0 +1,238 @@
...
+  });
+
+  @override
+  ConsumerState<ConsumerStatefulWidget> createState() => _SpecialFeaturePosterState();
+}
+
+class _SpecialFeaturePosterState extends ConsumerState<SpecialFeaturesRow> {
+  List<SpecialFeatureModel> get specialFeatures {
+    return widget.specialFeatures;
+  }
```
PartyDonut:
```suggestion
  List<SpecialFeatureModel> specialFeatures = widget.specialFeatures;
```
This can be moved to the build part of the widget. After that make the widget stateless as there is no state we need to maintain.

lib/screens/shared/media/special_features_row.dart
```diff
@@ -0,0 +1,238 @@
...
+import 'package:fladder/widgets/shared/item_actions.dart';
+import 'package:fladder/widgets/shared/modal_bottom_sheet.dart';
+import 'package:fladder/widgets/shared/status_card.dart';
+
+class SpecialFeaturesRow extends ConsumerStatefulWidget {
+  final List<SpecialFeatureModel> specialFeatures;
+  final String? label;
+  final ValueChanged<SpecialFeatureModel> playSpecialFeature;
+  final EdgeInsets contentPadding;
+  final Function(VoidCallback action, SpecialFeatureModel specialFeatureModel)? onSpecialFeatureTap;
```
PartyDonut:
```suggestion
  final Function(SpecialFeatureModel specialFeatureModel)? onSpecialFeatureTap;
```
Why does this take another VoidCallback as argument? I would suggest to only return the SpecialFeatureModel.

lib/screens/shared/media/special_features_row.dart
```diff
@@ -0,0 +1,238 @@
...
+                  widget.onSpecialFeatureTap?.call(
+                    () {
+                      specialFeature.play(context, ref);
+                    },
+                    specialFeature,
+                  );
+                }
+              : () {
+                  specialFeature.play(context, ref);
+                },
```
PartyDonut:
```suggestion
          onTap: widget.onSpecialFeatureTap != null
              ? () => widget.onSpecialFeatureTap?.call(
                    specialFeature,
                  )
              : () => specialFeature.play(context, ref),
```

Ok now we know why there was a action. But I'm not sure why that behaviour is there.
I would expect the function to be called if not null and we then supply the tapped specialFeature. Where we can then do whatever action needed in the details screens.

It's ok to fallback to a "default" action, but when the function has a override it should not still call the "default" action.

lib/providers/service_provider.dart
HachiHalation: I agree that we shouldnt swallow the exception here. But I do think the current callers should catch them and emit a log if there is a problem since in this context a failure should not stop the application.

lib/screens/shared/media/special_features_row.dart
HachiHalation: This behavior is consistent with the existing behavior from EpisodePosters widget. I don't really understand the concern, it is up to the caller to invoke the default action (in this case `play`) using the callback provided in onSpecialFeatureTap if they wish to. 

From what I understand from looking at the code, onSpecialFeatureTap provides a hook to callers to change the behavior of the tap action. onSpecialFeatureTap also provides a callback to the default action so that the caller can use it if they still want to perform the default actions during the customized logic. 

This all makes sense to me and is what was already implemented in EpisodePosters so I am keeping it like this for now. But please correct me if I am missing something and I will fix it.

PartyDonut: It's more that I like to explicitly know what "action" is. 
This makes the code just a little bit less readable because we have to dive into the widget itself to see what action actually does.

I know it also happens in the EpisodePosters but I don't like it there either, I should probably remove it there as well just did not get to that yet.

So it's fine to leave it as is, this works perfectly fine. But I would prefer the callback to be more explicit.

lib/providers/service_provider.dart
PartyDonut: Makes sense. Currently the handling of errors is a bit of a mess of my making.

Lets just leave it as is without the try/catch same as most of the other calls. I will have to come up with a proper solution when I can find the time.

HachiHalation: Sorry, I already moved the try catch before I saw this. Let me know if you want me to remove it anyways.

lib/screens/shared/media/special_features_row.dart
HachiHalation: Made some changes to hopefully make it more readable. If you still don't like it I will take your recommended solution instead.

PR #722, opened 2026-01-24
feat: improve CJK font rendering with chinese_font_library

### LINKED ISSUE #493: 🐛 The Simplified Chinese font display on the Windows version has errors.
### Description

In the latest nightly Windows version, the display of Simplified Chinese is abnormal, with fonts of different weights mixed together. I tested it on an Android - device and did not find this problem.

I will show you what this font display error roughly looks like in English.
**This** is a **simple. This is a simple**. This is a **simple.** 

Image

### Reproduction steps

```bash
1. Go to 'setting'
2. Change the display language to '简体中文（ZH）'
3. See error
```

### Screenshots

Image

### Logs

```bash

```

### Platform

Windows

### App Version

0.7.5Windows（284）

dongfengweixiao:
Integrate chinese_font_library to properly render Chinese, Japanese, and Korean characters across all platforms.

**I don’t have the environment to build and test.**

## Issue Being Fixed

Resolves #493

## Checklist

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -295,14 +295,27 @@ class _MainState extends ConsumerState<Main> with WindowListener, WidgetsBinding
     final language = ref.watch(clientSettingsProvider
         .select((value) => value.selectedLocale ?? WidgetsBinding.instance.platformDispatcher.locale));
     final scrollBehaviour = const MaterialScrollBehavior();
+    final isLinux = defaultTargetPlatform == TargetPlatform.linux;
     return DynamicColorBuilder(
       builder: (ColorScheme? lightDynamic, ColorScheme? darkDynamic) {
-        final lightTheme = themeColor == null
+        final baseLightTheme = themeColor == null
             ? FladderTheme.theme(lightDynamic ?? FladderTheme.defaultScheme(Brightness.light), schemeVariant)
             : FladderTheme.theme(themeColor.schemeLight, schemeVariant);
-        final darkTheme = (themeColor == null
+        final baseDarkTheme = (themeColor == null
             ? FladderTheme.theme(darkDynamic ?? FladderTheme.defaultScheme(Brightness.dark), schemeVariant)
             : FladderTheme.theme(themeColor.schemeDark, schemeVariant));
+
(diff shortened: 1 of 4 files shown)


dongfengweixiao:
|  before   | after  |
|  ----  | ----  |
| 屏幕截图 2026-01-25 125225  | 屏幕截图 2026-01-25 125653 |

lib/main.dart
```diff
@@ -295,14 +295,27 @@ class _MainState extends ConsumerState<Main> with WindowListener, WidgetsBinding
...
-        final lightTheme = themeColor == null
+        final baseLightTheme = themeColor == null
             ? FladderTheme.theme(lightDynamic ?? FladderTheme.defaultScheme(Brightness.light), schemeVariant)
             : FladderTheme.theme(themeColor.schemeLight, schemeVariant);
-        final darkTheme = (themeColor == null
+        final baseDarkTheme = (themeColor == null
             ? FladderTheme.theme(darkDynamic ?? FladderTheme.defaultScheme(Brightness.dark), schemeVariant)
             : FladderTheme.theme(themeColor.schemeDark, schemeVariant));
+
+        // Apply Chinese font for non-Linux platforms (Windows, macOS, Android, iOS)
```
PartyDonut, line 308: Why not apply the CJK font for all platforms?

pubspec.yaml
```diff
@@ -135,6 +135,9 @@ dependencies:
   dart_mappable: ^4.6.0
   flutter_native_splash: ^2.4.7
   macos_window_utils: ^1.9.0
+  
+  # Fix cjk font rendering issue on Windows
+  chinese_font_library: ^1.2.0
```
PartyDonut, line 140: I would prefer to just add additional fonts instead of relying on a external package for this.

dongfengweixiao: In that case, you would at least need to add fonts for Simplified Chinese, Traditional Chinese, and Japanese, which would lead to a significant increase in the size of the final build. Using chinese_font_library might be the solution with the least side effects.

PR #727, opened 2026-01-27
Add option to remove title bar in Linux and Windows

lkcv:
## Pull Request Description

Added a startup flag (--no-titlebar) to disable the title bar in Linux and Windows. This flag is ignored on macOS.

## Issue Being Fixed

This addresses this discussion:
https://github.com/DonutWare/Fladder/discussions/726

## Screenshots / Recordings

Here's a screenshot of Fladder in regular windowed mode (not htpc) without a title bar.
image

## Checklist

I've confirmed these changes work as expected on Linux.

### DIFF
--- a/lib/screens/shared/default_title_bar.dart
+++ b/lib/screens/shared/default_title_bar.dart
@@ -14,13 +14,20 @@ class DefaultTitleBar extends ConsumerStatefulWidget {
   final String? label;
   final double? height;
   final Brightness? brightness;
-  const DefaultTitleBar({this.height = defaultTitleBarHeight, this.label, this.brightness, super.key});
+  const DefaultTitleBar({
+    this.height = defaultTitleBarHeight,
+    this.label,
+    this.brightness,
+    super.key,
+  });
 
   @override
-  ConsumerState<ConsumerStatefulWidget> createState() => _DefaultTitleBarState();
+  ConsumerState<ConsumerStatefulWidget> createState() =>
+      _DefaultTitleBarState();
 }
 
-class _DefaultTitleBarState extends ConsumerState<DefaultTitleBar> with WindowListener {
+class _DefaultTitleBarState extends ConsumerState<DefaultTitleBar>
+    with WindowListener {
   bool hovering = false;
 
   @override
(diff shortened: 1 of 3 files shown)


PartyDonut (commented):
While the implementation is perfectly fine.

I'm thinking this might actually be better as a startup argument? 
Similar to how we have --htpc mode we could have --no-titlebar as a startup argument

lkcv:
Yeah, I can change it to a startup flag.

lkcv:
I got rid of the option and made it a startup flag instead.

lib/screens/shared/default_title_bar.dart
PartyDonut, line 1: Formatting is a bit all over the place. Could you please format the file using line length 120.

lib/screens/shared/default_title_bar.dart
```diff
@@ -37,12 +44,21 @@ class _DefaultTitleBarState extends ConsumerState<DefaultTitleBar> with WindowLi
...
+    final isOffline = ref.watch(
+      connectivityStatusProvider.select(
+        (value) => value == ConnectionState.offline,
+      ),
+    );
     final surfaceColor = theme.colorScheme.surface;
+    final noTitleBar =
+        (AdaptiveLayout.of(context).platform == TargetPlatform.linux ||
+                AdaptiveLayout.of(context).platform == TargetPlatform.windows) &&
+            ref.read(argumentsStateProvider).noTitleBar;
```
PartyDonut, line 61: Better to use a watch, it probably won't change during runtime. But better to be prepared.
```suggestion
            ref.watch(argumentsStateProvider.select((value) => value.noTitleBar));
```

lib/screens/shared/default_title_bar.dart
```diff
@@ -37,12 +44,21 @@ class _DefaultTitleBarState extends ConsumerState<DefaultTitleBar> with WindowLi
...
     final iconColor = theme.colorScheme.onSurface.withValues(alpha: 0.65);
-    final isOffline = ref.watch(connectivityStatusProvider.select((value) => value == ConnectionState.offline));
+    final isOffline = ref.watch(
+      connectivityStatusProvider.select(
+        (value) => value == ConnectionState.offline,
+      ),
+    );
     final surfaceColor = theme.colorScheme.surface;
+    final noTitleBar =
+        (AdaptiveLayout.of(context).platform == TargetPlatform.linux ||
```
PartyDonut, line 59: Let's move this logic to ArgumentsModel as a getter to clean it up. That way we only have to watch a single variable here.

You can also use defaultTargetPlatform, platform is only saved in AdaptiveLayout for convenience.

PR #728, opened 2026-01-28
feat: Use mouse side button to go back

lkcv:
## Pull Request Description

This makes it so the back mouse button (button on side of mouse that can go back to a previous window in a browser, etc.) can go back to the previous screen like the backspace key does.

## Issue Being Fixed

I didn't create a feature request for this since it's very minor.

## Checklist

Confirmed working on Linux, but should work on macOS and Windows too.


PartyDonut (approved):
Looks good to me thanks for helping out.

Next-up is getting the forward button to work :)

PR #736, opened 2026-02-03
feat: Implement Double-tap seek & press-and-hold playback speed

KeryWT:
## Pull Request Description

Add **two new features**:
- **press and hold to boost speed** to a configurable rate (spacebar on desktop and long press on mobile)
- **double tap to seek** forward or backward on mobile

## Issue Being Fixed

Resolves discussion #704 

## Screenshots / Recordings

https://github.com/user-attachments/assets/e4b17bcb-e519-49e5-96b2-b405c8b7bdb0

https://github.com/user-attachments/assets/b3180076-dffa-4102-9eae-6a6ff8d5db22

## Checklist

### DIFF
--- a/lib/routes/auto_router.dart
+++ b/lib/routes/auto_router.dart
@@ -16,6 +16,7 @@ const fullScreenRoutes = {
 const topBarNoBlurRoutes = {
   SettingsRoute.name,
   ControlPanelRoute.name,
+  DetailsRoute.name,
 };
 
 @AutoRouterConfig(replaceInRouteName: 'Screen|Page,Route')
--- a/lib/screens/details_screens/components/overview_header.dart
+++ b/lib/screens/details_screens/components/overview_header.dart
@@ -78,7 +78,7 @@ class OverviewHeader extends ConsumerWidget {
         );
 
     final fullHeight =
-        (MediaQuery.sizeOf(context).height - (MediaQuery.paddingOf(context).top + 150)).clamp(50, 1250).toDouble();
+        (MediaQuery.sizeOf(context).height - (MediaQuery.paddingOf(context).top + 50)).clamp(50, 1250).toDouble();
 
     final isPhone = AdaptiveLayout.viewSizeOf(context) == ViewSize.phone;
 
@@ -204,7 +204,7 @@ class OverviewHeader extends ConsumerWidget {
           mainAxisAlignment: MainAxisAlignment.end,
           crossAxisAlignment: crossAlignment,
           mainAxisSize: MainAxisSize.min,
-          spacing: 16,
+          spacing: 12,
           children: [
             if (!isPhone)
               Flexible(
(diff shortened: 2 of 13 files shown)


lib/screens/video_player/video_player_controls.dart
```diff
@@ -72,17 +79,63 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
...
+    
     return Listener(
       onPointerSignal: setVolume,
       child: InputHandler(
         autoFocus: true,
         keyMap: ref.watch(videoPlayerSettingsProvider.select((value) => value.currentShortcuts)),
         keyMapResult: _onKey,
+        onKeyEvent: isDesktop && speedBoostEnabled
+            ? (node, event) {
+                if (event.logicalKey == LogicalKeyboardKey.space) {
```
PartyDonut, line 104: This is a lot of logic, probably better to move it to a new function/class outside of the widget.
Also does this need the timer logic, would it not be easier to "activate" the speedup on `event is KeyRepeatEvent` and release it on `KeyUpEvent` for spacebar?

lib/screens/video_player/video_player_controls.dart
```diff
@@ -100,8 +153,49 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
...
+                            if (settings.enableSpeedBoost && isPlaying) {
+                              _activateSpeedBoost();
+                            }
+                          }
+                        : null,
+                    onLongPressEnd: initInputDevice == InputDevice.touch
+                        ? (details) {
+                            _deactivateSpeedBoost();
+                          }
+                        : null,
```
PartyDonut: Same here, we should probably think about moving the logic here.

lib/screens/video_player/components/video_player_seek_indicator.dart
```diff
@@ -11,6 +12,11 @@ import 'package:fladder/providers/video_player_provider.dart';
 import 'package:fladder/util/input_handler.dart';
 import 'package:fladder/util/localization_helper.dart';
 
+/// Provider to trigger the seek indicator from outside (e.g., double-tap).
+/// The value is the number of seconds (positive for forward, negative for backward).
+/// Set to null when not triggered externally.
+final seekIndicatorTriggerProvider = StateProvider<int?>((ref) => null);
```
PartyDonut: While this does work, if something goes wrong the state could remain on the UI even if the user has no pending seek actions.

Maybe it would be better to create a controller and pass that through from the `video_player_controls.dart` to this widget. That way the GestureDetector could just call seekBack/seekForward in this widget using the controller without having to rely on a provider.

PartyDonut (changes requested):
Thanks for having a go at this, while the code looks good. Left a few requests, let me know if they aren't clear.

Also something else to consider. My idea for implementing the Gestures was to have predefined areas/actions the user could assign shortcuts to, in a similar fashion to how desktop users can assign keys to a action. 

So while I'm more then happy to merge this after the requested changes, there is a possibility that I will rework some of this when I get to the gestures implementation. Just wanted to give you a heads up 😃 .

lib/screens/video_player/video_player_controls.dart
KeryWT: You're absolutely right, yup. I've simplified the logic by using the KeyRepeatEvent.
The play/pause is now triggered by the KeyUpEvent in order to avoid stopping the playback when holding spacebar for speedBoost.

lib/screens/video_player/video_player_controls.dart
KeryWT: Done!

lib/screens/video_player/components/video_player_seek_indicator.dart
KeryWT: Alright, I tried giving it a go.
Let me know what you think 😊

PR #754, opened 2026-02-09
Remove duplicate play button

lkcv:
## Pull Request Description

Removes the big desktop-only play button in the middle of the screen that persists regardless of mouse panning.
There's already a play/pause button at the bottom control bar, so now it has tooltips depending on the playback state (Pause video and Resume video).

## Issue Being Fixed

I haven't created an issue for it yet, but will if needed.

## Screenshots / Recordings
image

## Checklist
- Tested on Linux, but this should work on macOS and Windows too

### DIFF
--- a/lib/screens/video_player/video_player_controls.dart
+++ b/lib/screens/video_player/video_player_controls.dart
@@ -336,6 +336,7 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
                     icon: Icon(
                       mediaPlayback.playing ? IconsaxPlusBold.pause : IconsaxPlusBold.play,
                     ),
+                    tooltip: mediaPlayback.playing ? "Pause video" : "Resume video",
                   ),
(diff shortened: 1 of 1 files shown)


talynone:
I kind of like the play button in the middle. If the video is paused windowed and there's no mouse focus on the player it's a good indicator the player is in a pause state. Also more UI friendly to older folks when teaching them how to use app.

lkcv:
@talynone, I was trying to make it more like the mobile apps since those only show the single Play/Pause button at the bottom whether the screen is touched or not during playback. I also don't like the icon in the middle 😆

I can add it behind an option, but I'll wait until PartyDonut has a chance to review it.

PartyDonut:
Like @talynone mentioned, I like having a pause button in the middle as a indicator that it's actually paused. 

So closing this for now. Feel free to open a discussion however for any UI/UX ideas.

PR #643, opened 2025-12-04
fix: Parsing non-ascii domains

kikibase:
- added puny coder dependency
-  non-ascii domains can now be resolved

## Pull Request Description

just changed api_provider.dart to use punycoder to encode non-ascii characters in domain names

## Issue Being Fixed

Resolves #640

## Screenshots / Recordings

no visual change

## Checklist

- [x] If a new package was added, did you ensure it works for all supported platforms? Is the package well maintained
- [x] Check that any changes are related to the issue at hand.

PR #773, opened 2026-02-18
fix: Format `Locale` objects scriptcode correctly

### LINKED ISSUE #641: 🐛 After selecting the Traditional Chinese language, restarting will revert to the Simplified Chinese language
### Description

After adjusting the settings to display language to Traditional Chinese, it will be changed back to Simplified Chinese upon restart

### Reproduction steps

```bash
1. Go to 'Setting'
2. Click on 'Display language'
3. Select '繁體中文(ZH)'
4. Restart
```

### Screenshots

_No response_

### Logs

```bash

```

### Platform

Windows

### App Version

v0.8.0

### Jellyfin server

10.10.7

d0gkiller87:
## Pull Request Description

The existing logic for:
- Displaying settings: https://github.com/DonutWare/Fladder/blob/v0.9.0/lib/util/localization_helper.dart#L62
- Saving/loading settings: https://github.com/DonutWare/Fladder/blob/v0.9.0/lib/models/settings/client_settings_model.dart#L127

Doesn't take `Locale.scriptCode` into account during formatting/serialization, turning language codes like `zh_Hant` into `zh`. This PR addresses the issue and formats/parses language codes according to the pattern of BCP 47. (i.e. `EN_US`→`en_US`, `ZH_HANT`→`zh_Hant`)

## Issue Being Fixed

Resolves #641

## Checklist

### DIFF
--- a/lib/util/localization_helper.dart
+++ b/lib/util/localization_helper.dart
@@ -64,12 +64,15 @@ extension LocaleDisplayCodeExtension on Locale {
 
     buffer.write(languageCode.toLowerCase());
 
-    if (scriptCode != null && scriptCode!.isNotEmpty) {
-      buffer.write('_${scriptCode![0].toUpperCase()}${scriptCode!.substring(1).toLowerCase()}');
+    final scriptCode = this.scriptCode;
+    final countryCode = this.countryCode;
+
+    if (scriptCode != null && scriptCode.isNotEmpty) {
+      buffer.write('_${scriptCode[0].toUpperCase()}${scriptCode.substring(1).toLowerCase()}');
     }
 
-    if (countryCode != null && countryCode!.isNotEmpty) {
-      buffer.write('_${countryCode!.toUpperCase()}');
+    if (countryCode != null && countryCode.isNotEmpty) {
+      buffer.write('_${countryCode.toUpperCase()}');
     }
 
     return buffer.toString();


d0gkiller87:
One thing to note: since the original implementation already saves language codes in BCP 47's format, existing users' selected language won't be affected by this change as tested by migrating from version 9.0.

lib/util/localization_helper.dart
```diff
@@ -60,9 +60,19 @@ class _LocalizationContextWrapperState extends ConsumerState<LocalizationContext
...
+      buffer.write('_${scriptCode![0].toUpperCase()}${scriptCode!.substring(1).toLowerCase()}');
+    }
+
+    if (countryCode != null && countryCode!.isNotEmpty) {
+      buffer.write('_${countryCode!.toUpperCase()}');
+    }
+
+    return buffer.toString();
   }
 }
```
PartyDonut, line 80: Small suggestion to get rid of the bang operators

```suggestion
extension LocaleDisplayCodeExtension on Locale {
  String toDisplayCode() {
    final buffer = StringBuffer();

    buffer.write(languageCode.toLowerCase());

    final scriptCode = this.scriptCode;
    final countryCode = this.countryCode;

    if (scriptCode != null && scriptCode.isNotEmpty) {
      buffer.write('_${scriptCode[0].toUpperCase()}${scriptCode.substring(1).toLowerCase()}');
    }

    if (countryCode != null && countryCode.isNotEmpty) {
      buffer.write('_${countryCode.toUpperCase()}');
    }

    return buffer.toString();
  }
}
```

PartyDonut (changes requested):
Looks good to me, small change request just to make sure.

d0gkiller87:
Another thing before merging:
Not sure how `localeResolutionCallback` (pointed out by @kikibase @ https://github.com/DonutWare/Fladder/pull/643/changes/a0182ef259517c525b3a0b3768b7c32d128ac462) affects the locale loading process. In my test runs, taking care of `scriptCode` already mitigated the issue. Can someone check what `localeResolutionCallback` actually does in Fladder?

d0gkiller87:
> Another thing before merging: Not sure how `localeResolutionCallback` (pointed out by @kikibase @ [a0182ef](https://github.com/DonutWare/Fladder/commit/a0182ef259517c525b3a0b3768b7c32d128ac462)) affects the locale loading process. In my test runs, taking care of `scriptCode` already mitigated the issue. Can someone check what `localeResolutionCallback` actually does in Fladder?

Never mind, I understand it now.

For reference, the code in question:
https://github.com/DonutWare/Fladder/blob/v0.9.0/lib/main.dart#L321

It defines both `locale` and `localeResolutionCallback`. According to the Flutter documentation: https://api.flutter.dev/flutter/widgets/WidgetsApp/locale.html

> `localeResolutionCallback`, which can override the default supportedLocales matching algorithm.

And given this block:

```dart
if (supportedLocales.contains(locale)) {
  return locale;
}
```

the `(l) => l.languageCode == locale.languageCode` check effectively serves as a fallback. It only runs when `locale` is not found in `supportedLocales`.

If we were changing it to `(l) => l == locale`, it would likely never match, since `supportedLocales.contains(locale)` has already failed at that point ...

If everything checks out, please feel free to merge the PR, thanks!

PR #777, opened 2026-02-20
fix: libass on android

frdmmm:
## Pull Request Description

let the app to render srt subtitles on android, and libass to render ass or ssa

## Checklist


PartyDonut (approved):
Thanks looks good!

PR #797, opened 2026-02-28
Fix/disable brightness desktop

Julien9969:
## Pull Request Description
If the screen supports software-based brightness control, the desktop application automatically sets the brightness to minimum when the application loses focus.

As it seem unused, i remove the brightness control on desktop
## Checklist

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -15,6 +15,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:macos_window_utils/window_manipulator.dart';
 import 'package:package_info_plus/package_info_plus.dart';
 import 'package:path_provider/path_provider.dart';
+import 'package:screen_brightness/screen_brightness.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:smtc_windows/smtc_windows.dart' if (dart.library.html) 'package:fladder/stubs/web/smtc_web.dart';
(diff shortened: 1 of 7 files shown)


PartyDonut (commented):
What about users on Windows tablets. Or similar form factors on a desktop OS.

Julien9969:
> What windows users with tablets?

You mean, what for tablette windows user or am i a windows tablette user ?
My issue happen with a desktop PC and samsung external screen.

PartyDonut:
Sorry, too much typing too little thinking on my end 😅 .

But yes the question was for users on windows tablets. The brightness control is probably something that is useful in that context?

PartyDonut:
Maybe add a check for pointer devices? 

So enable brightness control for android, iOS and when the user is using either touch or dpad controls?

Julien9969:
> Sorry, too much typing too little thinking on my end 😅 .
> 
> But yes the question was for users on windows tablets. The brightness control is probably something that is useful in that context?

I can’t manage to stop the brightness control from messing with my monitor’s brightness, except by completely removing it. The issue doesn’t seem to happen often, but I’m also not sure there are many Windows tablet users.
It’s a trade-off, and I don’t know what the best solution is.

Julien9969:
> Maybe add a check for pointer devices?
> 
> So enable brightness control for android, iOS and when the user is using either touch or dpad controls?

I can try this solution, for sure.

PR #803, opened 2026-03-04
chore: Add workflow step to update metainfo.xml for releases

schembriaiden:
## Pull Request Description

This PR updates the build workflow to include an update for the metainfo.xml file when creating a new release.

The updated metainfo.xml file is then committed back into the develop branch so that the new update is reflected.

## Issue Being Fixed

Resolves #issue-number

## Checklist


.github/workflows/build.yml
```diff
@@ -381,10 +381,71 @@ jobs:
             Fladder_x86_64.AppImage
             Fladder_x86_64.AppImage.zsync
 
+  prepare-metainfo:
+    runs-on: ubuntu-latest
+    needs: [fetch-info]
+    if: needs.fetch-info.outputs.build_type == 'release'
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6.0.2
```
PartyDonut: Let's use the same checkout version as the other steps. Or update the other steps to this version

PartyDonut (changes requested):

.github/workflows/build.yml
schembriaiden: Done

PartyDonut (approved):

PR #812, opened 2026-03-06
feat: Seerr QC login, password-login hide, auto-URL scheme, new ENV vars | fix: Set-Cookie

v3DJG6GL:
## Pull Request Description

- This PR adds QuickConnect authentication support to the login flow Seerr (see PR https://github.com/seerr-team/seerr/pull/2212), which enables passwordless authentication for Seerr.
- Furthermore, it adds several smaller UX changes/improvements: Direct Jellyfin QC settings link, copy-paste QC code, a hide PW menu option and some other small UX changes.
- I've additionally added a fix for the issue mentioned here: https://github.com/DonutWare/Fladder/issues/710#issuecomment-4012345931
- Additionally, I've experimented with setting custom headers through ENV vars. Although I am not sure whether the additional code changes are really worth it tbh...

---
### Seerr QuickConnect Integration (All Plattforns)
- Add QuickConnect API endpoints and models (`lib/seerr/seerr_models.dart`)
- Add QuickConnect service methods (`lib/providers/seerr_service_provider.dart`)
- Add QuickConnect tab to Seerr connection dialog
- Auto-restart QC polling after in-app browser closes
- Add 5-minute QC polling timeout
- Clear QC secrets on dispose

### UI / UX (All Plattforns)
- Add "Open Jellyfin QuickConnect" link to QuickConnect modals
- Copy QuickConnect code to clipboard on tap **(Addresses #653)**
- Auto-add URL scheme for Jellyfin and Seerr URLs (try https first, then http)
- Add `hidePasswordLogin` client setting to hide password fields
- Rename "Local Server URL" to "Local Jellyfin URL"

### Docker: New Environment Variables
- **`SEERR_URL`**: Pre-fills and locks the Seerr URL in settings **(Addresses #688)**
- **`HIDE_PASSWORD_LOGIN`**: Hides password fields, showing only QuickConnect **(_partially_ Addresses #652)**
- **`SEERR_CUSTOM_HEADERS`**: JSON object of headers injected server-side for all users.
    - Headers are injected server-side via an nginx reverse proxy (`/seerr-proxy/`) so header values should never reach the client.

---

## Issue Being Fixed

- **Seerr login on web (Fixes #710):**
  - Credential and QC login on web failed with "No session cookie returned by server" because browsers never expose `Set-Cookie` to JavaScript
  - Use `BrowserClient` with `withCredentials = true` so the browser manages session cookies natively
  - Skip manual cookie extraction on web; use sentinel value to signal browser-managed auth

---

[...]

PR #817, opened 2026-03-08
feat: Implement Seerr url and header

PartyDonut:
## Pull Request Description

Implement seerr header and base url #688

### DIFF
--- a/lib/util/fladder_config.dart
+++ b/lib/util/fladder_config.dart
@@ -6,12 +6,29 @@ class FladderConfig {
   static set baseUrl(String? value) => _instance._baseUrl = value;
   String? _baseUrl;
 
+  static String? get seerrBaseUrl => _instance._seerrBaseUrl;
+  static set seerrBaseUrl(String? value) => _instance._seerrBaseUrl = value;
+  String? _seerrBaseUrl;
+
+  static Map<String, String>? get seerrHeader => _instance._seerrHeader;
+  static set seerrHeader(Map<String, String>? value) => _instance._seerrHeader = value;
+  Map<String, String>? _seerrHeader;
+
   static void fromJson(Map<String, dynamic> json) => _instance = FladderConfig._fromJson(json);
 
   factory FladderConfig._fromJson(Map<String, dynamic> json) {
     final config = FladderConfig._();
     final newUrl = json['baseUrl'] as String?;
+    final newSeerrUrl = json['seerrBaseUrl'] as String?;
+    final newSeerrHeader = json['seerrHeader'];
+
     config._baseUrl = newUrl?.isEmpty == true ? null : newUrl;
+    config._seerrBaseUrl = newSeerrUrl?.isEmpty == true ? null : newSeerrUrl;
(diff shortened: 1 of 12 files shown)


v3DJG6GL:
@PartyDonut Are you aware that this implementation will expose `SEERR_HEADER` value at `/assets/config/config.json` to everyone who has access to Fladder?
I've initially made the same implementation, but then switched to an internal nginx reverse proxy to inject headers server-side. This has the additional advantage that there is no need to configure CORS headers at external reverse proxy...

lib/util/fladder_config.dart
```diff
@@ -6,12 +6,29 @@ class FladderConfig {
...
 
   factory FladderConfig._fromJson(Map<String, dynamic> json) {
     final config = FladderConfig._();
     final newUrl = json['baseUrl'] as String?;
+    final newSeerrUrl = json['seerrBaseUrl'] as String?;
+    final newSeerrHeader = json['seerrHeader'];
+
     config._baseUrl = newUrl?.isEmpty == true ? null : newUrl;
+    config._seerrBaseUrl = newSeerrUrl?.isEmpty == true ? null : newSeerrUrl;
+    if (newSeerrHeader is Map<String, String>) {
```
v3DJG6GL, line 27: if (newSeerrHeader is Map) {
      final parsed = newSeerrHeader.map((key, value) => MapEntry(key.toString(), value.toString()));

v3DJG6GL: sorry for spamming, I really don't want fill your inbox 🙈

I found that current `SEERR_HEADER` implementation is not working for me:
I've set
```
SEERR_HEADER='{"Authorization":"Basic ..."}'
```
However, this header is ignored and never added to requests towards Seerr.

I suppose it's because the config JSON is parsed by `jsonDecode()` which returns `Map<String, dynamic>`, but the type check in `fladder_config.dart:27` uses `is Map<String, String>` which then always fails (dynamic≠String)....?

And therefore, the header is never loaded into `FladderConfig.seerrHeader`.

PartyDonut:
> @PartyDonut Are you aware that this implementation will expose `SEERR_HEADER` value at `/assets/config/config.json` to everyone who has access to Fladder? I've initially made the same implementation, but then switched to an internal nginx reverse proxy to inject headers server-side. This has the additional advantage that there is no need to configure CORS headers at external reverse proxy...

Yes, I guess my hope was users would not add this to public facing servers. But I'm pretty sure users are going to do this anyway.

I'm not sure there is a "safe" way to add those headers without exposing it so long as the call is made inside of a browser. It will eventually show those headers in any network calls.

v3DJG6GL:
> Yes, I guess my hope was users would not add this to public facing servers. But I'm pretty sure users are going to do this anyway.

haha - yep, it might probably happen... 🙈

> I'm not sure there is a "safe" way to add those headers without exposing it so long as the call is made inside of a browser. It will eventually show those headers in any network calls.

With my implementation from PR #812, I've added nginx as proxy. As far as I could tell/verify, no header were exposed to clients in any way.
But yeah, it adds a bit more complexity...

PR #821, opened 2026-03-08
feat: Add Fastlane changelog generation and metadata for F-Droid

schembriaiden:
## Pull Request Description

This PR adds the Fastlane structure recommended by F-Droid to release the app where we are able to control the metadata ourselves instead of opening a PR on the F-Droid repo every time we want to change the metadata.

What this does is:

- Has a workflow to create the release notes to be shown on F-Droid
- Has a workflow to validate the Fastlane folder (can be removed if you think its extra)
- Copied the 512x512 icon
- Copied the phone marketing photos. (I was thinking maybe instead of copying them we just move them to the Fastlane and we control them from there as it seems a bit extra having them in both places, but you let me know what you want to do.)

Fastlane has a lot more use cases which I did not look into but maybe you would want to:

https://docs.flutter.dev/deployment/cd

Additional context for this specific PR:

https://gitlab.com/-/snippets/1895688
https://docs.fastlane.tools/
https://github.com/ashutoshgngwr/validate-fastlane-supply-metadata
https://gitlab.com/schembriaiden/fdroiddata/-/blob/nl.jknaapen.fladder/metadata/nl.jknaapen.fladder.yml

image

## Checklist


.github/workflows/build.yml
```diff
@@ -693,6 +693,58 @@ jobs:
...
+          # Strip markdown: remove PR links, @mentions, image URLs, backticks, excess whitespace
+          CLEAN=$(echo "$RELEASE_BODY" \
+            | sed 's/\[#[0-9]*\]([^)]*)//g' \
+            | sed 's/ in \[#[0-9]*\]([^)]*)//g' \
+            | sed 's/by @[A-Za-z0-9_-]*//g' \
+            | sed 's/!?\[.*\]([^)]*)//g' \
+            | sed 's/`//g' \
+            | sed '/^---/d' \
+            | sed 's/[[:space:]]\{2,\}/ /g' \
+            | sed '/^[[:space:]]*$/d')
```
PartyDonut: Not going to lie this seems like magic to me so no idea if the regex are correct. Would it be acceptable for f-droid to simply link to the changelog?

There is also the possibility that quite a few changes are not related to the Android build. Should we filter those out?

PartyDonut (commented):
Looks good to me, left one small comment/question.

.github/workflows/build.yml
schembriaiden: It should be fine to just link it to the Github Changelog so I did that

PartyDonut (approved):
Thanks for the quick update 👍🏼

Issue #710, opened 2026-01-14
Jellyseerr local login fails from web (proxmox helper script); works fine from android


tismondo:
### Description

I tried connecting my Fladder instance to Jellyseer via local IP using the web interface and keep getting the below error.  The jellyseerr container is running and the same settings configured in the android app work fine, so I'm not sure if there's a bug here or user error.

I concede it's possible something specific about the LXC setup in proxmox is confusing things, but wanted to flag it.

<img width="1070" height="224" alt="Image" src="https://github.com/user-attachments/assets/3aaccac2-fa04-45f7-b9cb-d23ec16e6c13" />

### Reproduction steps

```bash
1. Go to 'Settings > Profile > Advnaced'
2. Click on 'Seerr'
3. Input my working local IP (again - tested to be working via the Android app)
4. See error
```

### Logs

```bash

```

### Platform

Web

### App Version

0.9.0

### Jellyfin server

Proxmox LXC

divemasterjm:
+1

Keyplayerone:
I am also receiving this specific error message. I discovered Fladder via Proxmox Helper scripts and find it very useful. However, the integration of Jellyseer did not work. Since I suspected a problem with helper scripts, I set up a separate Docker container - unfortunately, I encountered the exact same issue here.

roormonger:
So from the digging I did, this is apparently a CORS problem. Using an API key will get it working in the web ui. Might be the only clean way...

divemasterjm:
i've tried 3 methods, all give same error

simon-dmsn:
Have you tried putting https (or http) at the beginning of the URL ? I had the same issue without and got it solved by adding https in front of my seer URL

If it can help :)

tismondo:
> Have you tried putting https (or http) at the beginning of the URL ? I had the same issue without and got it solved by adding https in front of my seer URL

My Jellyseerr is local-only and not behind tls but yes, I've tried http and https.

PartyDonut:
> So from the digging I did, this is apparently a CORS problem. Using an API key will get it working in the web ui. Might be the only clean way...

Yup this is a CORS error. This is not something I can fix in the Fladder web hosted version, you need Seerr to allow the connections comming from the Fladder host.

Wunderharke:
~~So should this be raised in the seerr repo? Or is there already a configuration for this?~~
I added CORS headers manually at my reverse proxy, but now I get "No session cookie returned by server".
[...]

PR #826, opened 2026-03-10
fix: Seerr login browser cookie handling on container deployment

v3DJG6GL:
## Pull Request Description
- Seerr connection fails on container deploymenth with _No session cookie returned by server._
- Browsers process the `set-cookie` response header internally but never expose it to JavaScript. Fladder can't capture the session cookie and treats login as failed. Much better described by @Wunderharke: https://github.com/DonutWare/Fladder/issues/710#issuecomment-4012345931

- This fix uses a conditional-import `BrowserClient` with `withCredentials: true` so the browser manages session cookies on its own. Since Fladder can't see whether a cookie was set, a `__browser_managed__` sentinel is stored on web instead,
  and `_authHeaders` skips the manual Cookie header since the browser should already attach it.
- This feature was part of PR #812

_NOTE: Cross-origin setups still require CORS headers on the Seerr side_

## Testing Build
```
ghcr.io/v3djg6gl/fladder:test-seerr-web-login-fix
```

## Issue Being Fixed

- https://github.com/DonutWare/Fladder/issues/710#issuecomment-4012345931

## Screenshots / Recordings

N/A

### AI Disclosure
*Developed with assistance from Claude Code (Opus 4.6). The AI helped me identify the related files, code parts and proposed possible code modifications and fixes.*
*I've manually reviewed, cleaned, refactored and hardened the code the best I could, but since my Dart capabilities are not that good, I see this PR more as a proposal/idea.*
***A Proper code review is required in any case...***

## Checklist

### DIFF
--- a/lib/providers/seerr_api_provider.dart
+++ b/lib/providers/seerr_api_provider.dart
@@ -11,6 +11,8 @@ import 'package:fladder/providers/user_provider.dart';
 import 'package:fladder/seerr/seerr_chopper_service.dart';
 import 'package:fladder/seerr/seerr_json_converter.dart';
 import 'package:fladder/util/fladder_config.dart';
+import 'package:fladder/util/seerr_http_client.dart'
+    if (dart.library.html) 'package:fladder/util/seerr_http_client_web.dart';
 
 part 'seerr_api_provider.g.dart';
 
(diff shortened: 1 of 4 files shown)


PartyDonut (approved):
LGTM, Thanks.

PR #833, opened 2026-03-10
chore: Remove seerrHeader from config.json

PartyDonut:
## Pull Request Description

Remove seerrHeaders from config.json

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -141,17 +141,15 @@ Run `docker-compose up -d` to start the container. It will be available on `http
 > [!TIP]
 > We recommend changing the `BASE_URL` environment variable to the URL you use to access Jellyfin, as this will skip entering it when you load the web UI.
 
-You can also preconfigure Seerr with these environment variables:
+You can also preconfigure Seerr with this environment variable:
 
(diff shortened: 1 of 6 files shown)


v3DJG6GL:
@PartyDonut 
I've seen you reverted Seerr header implementation through ENV value.

I did a new look at it and implemented it with a different approach (similar as I did it once before):
https://github.com/DonutWare/Fladder/compare/develop...v3DJG6GL:Fladder:feat/seerr-custom-headers
AFAIK, that should not reveal configured header values to end users.

PartyDonut:
> @PartyDonut I've seen you reverted Seerr header implementation through ENV value.
> 
> I did a new look at it and implemented it with a different approach (similar as I did it once before): [develop...v3DJG6GL:Fladder:feat/seerr-custom-headers](https://github.com/DonutWare/Fladder/compare/develop...v3DJG6GL:Fladder:feat/seerr-custom-headers) AFAIK, that should not reveal configured header values to end users.

Have not tested it. But the browser would still leak it when the api call is made right? If not then by all means a PR would be great 😄

PR #851, opened 2026-03-18
fix: Android native player trickplay incorrect offsetting

matt-hb:
## Pull Request Description

Due to a maximum texture size set by Coil (4096 pixels in either dimension on my phone), my trickplay tiles were being downscaled to 4096, without the `trickPlayModel` knowing about it, causing some incorrect display on the timeline. This PR rescales the display offset and size to match this downscaled bitmap.

Also fixed trickplay images being stretched (ugly) when the images are not 16:9, which is true in many movies especially.

## Checklist


PartyDonut (approved):
Looks good to me, thanks for fixing this.

PR #852, opened 2026-03-18
fix: Android native player's chapter markers turn into huge circles when seeking

matt-hb:
## Pull Request Description

Screenshots explain better, the PR makes the chapter markers similar to the thumb when seeking, making it more uniform (also upped the opacity on upcoming chapter markers, they were hardly visible before).

### Before:
image

### After:
image


PartyDonut (approved):
Looks good, never noticed this problem before though which is weird.

Probably the aspectRatio acting up under certain conditions?

On a side note the progressbar hiding like this is by all means a uninted side effect on my part. But it works and actually makes the trickplay items more visible.

PR #866, opened 2026-03-21
feat(player): add customizable aspect ratio override

Vigno04:
## Pull Request Description

Added customizable aspect ration override, the override was inplemented on the player but it could be easily moved to the option panel if needed, usefull for content where the video was encoded differntly, for example the new one piece tv show where the video is 16:9 but the content is 2:1 so now there are'nt any way to watch it at the macimum size with no content cut

## Screenshots / Recordings

immagine
immagine

### DIFF
--- a/lib/models/settings/client_settings_model.dart
+++ b/lib/models/settings/client_settings_model.dart
@@ -108,7 +108,7 @@ abstract class ClientSettingsModel with _$ClientSettingsModel {
   }
 
   Future<String?> getSavePath() async {
-    if (kIsWeb && syncPath == null) return null;
+    if (kIsWeb) return syncPath;
     if (Platform.isMacOS) {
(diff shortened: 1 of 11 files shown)


PartyDonut:
Thanks for implementing this, while the code itself looks good. This seems a bit out of scope for what Fladder is supposed to handle.

For things like this I would recommend people to crop the video itself. Feel free to open a discussion and check if there are more people with a similar problem. 

It also seems to crop out subtitles which is probably not the intended effect. To implement this we should probably look at implementing it at the player level not by cropping the output of said player.

Vigno04:
Ok, will open a discussion, for the subtitle I can look into it if needed

Vigno04:
Also the problem with cropping the source especially a 4k HDR stream is that it could not be done lossless and also it takes a ton of times.

PR #876, opened 2026-03-27
fix: Slideshow banner thumb incorrect position

matt-hb:
## Pull Request Description

Slideshow banner's slider logic to switch between items by dragging is more natural this way, whichever item is closest instead of the one on the left of the cursor (record for tiniest PR?).


PartyDonut (approved):
Thanks did not even notice this. Yes smallest PR ever exactly how I like them 😄

PR #884, opened 2026-03-29
Native Dolby Vision & High-Quality Audio Support (Windows), with external player

NEVARLeVrai:
## Pull Request Description

This Pull Request introduces native **Dolby Vision (DV)** playback and advanced audio codec support (Atmos, DTS:X, 5.1 passthrough) for Fladder on Windows by integrating **Energy Media Player** as an optional external player.

### Key Features
- **Intelligent DV Detection**: Identifies Dolby Vision content (DOVI, HDR10, HLG hybrids) using Jellyfin metadata.
- **External Player Bridge**: Robust `energyplayer:` protocol launch with synced playback position and subtitles.
- **User Preference System**: Selection dialog with "Remember my choice" persistence.
- **Localized Guidance**: Integrated instructions for Energy Player configuration (Passthrough & DV decoding).

### Technical Details & Fixes
- Resolved `RepeatMode` namespace collision.
- Implemented URL normalization to prevent double slashes and hardened URI encoding for cross-device compatibility.

###  AI Contribution Note
- Architected and implemented with the assistance of the **Antigravity AI agent**. 
- AI helped with complex model mapping, troubleshooting Windows build failures, and localizations. A manual code review is - recommended.

## Issue Being Fixed

- **Feature Request**: Support for native Dolby Vision on Windows.
- **Problem**: Internal players (MPV/Media Kit) often lack bit-perfect DV rendering on Windows, resulting in washed-out colors.

## Screenshots / Recordings

image
image
image

## Checklist

- [x] If a new package was added, did you ensure it works for all supported platforms? (No new packages added)

### Detailed Technical Note (Rationale & Implementation)

**The Problem: Dolby Vision on Windows**
Standard internal players in Fladder (based on MPV or Media Kit) often struggle with native Dolby Vision (profiles 5, 7, and 8). On many Windows systems, this results in "washed out" colors because the player attempts to map the DV metadata into SDR/HDR10 incorrectly, or fails to trigger the OS-level DV rendering pipeline.

**The Solution: Energy Media Player Integration**
Energy Media Player is a specialized player for Windows that leverages native **WinRT/UWP Media Foundation** components and **HEVC Video Extensions**. By handing off the stream to this player:
1. **Bit-Perfect Rendering**: It utilizes the official Windows Dolby Vision extensions to guarantee accurate color mapping and dynamic metadata handling.
2. **High-End Audio**: It provides robust support for **Audio Passthrough** (Atmos, DTS:X, 7.1) which can be challenging to configure reliably in cross-platform internal players.
3. **Optimized Streaming**: The integration uses a custom URI protocol (`energyplayer:`) allowing Fladder to pass the authenticated Jellyfin stream URL, playback position, and subtitle tracks in a single seamless handoff.

**User Impact**
This feature was driven by a personal need for a cinematic experience on PC hardware. It allows users with DV-compatible monitors/TVs to finally enjoy their library without compromise, choosing between the flexibility of the internal player for standard content and the specialized power of Energy Player for high-bitrate DV material.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -2570,5 +2570,17 @@
   "@groupedFoldersDescription": {},
   "downloadTranscoded": "Download transcoded",
   "downloadOriginal": "Download original",
-  "today": "Today"
+  "today": "Today",
+  "dvPlayerSelectionTitle": "Dolby Vision Player",
+  "dvPlayerSelectionDesc": "Choose how to play Dolby Vision on Windows. Energy Player is required for native vision and high-quality audio passthrough (Atmos, 5.1).",
+  "dvPlayerAsk": "Always ask",
+  "dvPlayerDisabled": "Disabled (Internal player)",
+  "dvPlayerEnergyPlayer": "Energy Player (External)",
+  "dvPlayerDialogTitle": "Play with...",
+  "dvPlayerDialogDesc": "This content is in native Dolby Vision. Use Energy Player for accurate HDR colors and advanced audio support (Atmos/5.1 passthrough). Internal playback may have dull colors or limited audio codec support.",
+  "dvRememberChoice": "Remember my choice",
+  "dvPlayerNotInstalled": "Energy Player is not installed. Would you like to download it from the Microsoft Store?",
+  "dvPlayerGetFromStore": "Get from Store",
+  "dvEnableInstruction": "Configuration for native Dolby Vision and Audio:\n• Enable 'Allow Dolby Vision decoding' in Energy Player 'Video/Audio settings'.\n• Enable relevant 'Passthrough' options for Atmos/5.1 support.\n• Ensure HDR is enabled in Windows Display settings.",
+  "dvSyncWarning": "Note: Playback state is saved within Energy Player only and will not sync back to Fladder."
 }
--- a/lib/models/items/media_streams_model.dart
+++ b/lib/models/items/media_streams_model.dart
@@ -71,6 +71,8 @@ class MediaStreamsModel {
 
   String? get mediaInfoTag => '${displayProfile?.value} ${resolution?.value}';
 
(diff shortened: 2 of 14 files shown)


NEVARLeVrai:
I hope this helps you and others who want to enjoy DV on Windows, using your app.

There have been some issues reported regarding Copilot AI. While I haven't used Copilot AI myself, I understand there's a need to accept changes related to it. I'll let it up to you to review these changes.

And have a great Day !

PartyDonut:
Hi sorry for the late response, I have been thinking about the implementation.

I feel in it's current state it relies to heavily on a specific player.

Maybe a better solution would be to specify a few presets for launching external players that the user can then edit manually as well? If that's even possible.

NEVARLeVrai:
> Hi sorry for the late response, I have been thinking about the implementation.
> 
> I feel in it's current state it relies to heavily on a specific player.
> 
> Maybe a better solution would be to specify a few presets for launching external players that the user can then edit manually as well? If that's even possible.

I’m not sure, since Energy Player relies on some specific launch arguments. I’m also not certain that other players even support launching directly from a streaming URL.

So far, it’s the only player I’ve found that properly supports both Dolby Atmos and Dolby Vision while also handling streaming URLs correctly.

PartyDonut:
For now lets simplify it down to just external players. I think most players launch in a similar fashion using a URL with arguments for Android at least.

Lets re-use the base class you created `ExternalPlayerHelper` (maybe rename to "ExternalPlayer"?) and have all other external players base of that so they all share base functions. 

Probably for now we would need the following functions

"launch(ItemModel item)" -> returns boolean if launch was successful.  Also what do we show in Fladder when the external player is opened for desktop platforms?
"installed" -> returns boolean let the external player specify what check is done
"openStore" -> 

While I do understand the want for the dynamic switching lets keep that out for now. 

To determine what ExternalPlayer is available for which platform I would suggest a similar implementation like the current video player backend options.

Would be great to implement this for mobile platforms as well or at least keep them in mind.

PR #895, opened 2026-04-01
feat: Implement edge gesture controls for brightness and volume

schembriaiden:
## Pull Request Description

This pull request adds support for edge gestures in the video player settings, allowing users to control brightness and volume by sliding on the screen edges. It introduces new settings for enabling/disabling and swapping edge gestures, updates the localization files with relevant strings, and improves synchronization between the app's internal volume and the system volume on mobile platforms.

**Edge Gestures Feature:**

- Added new settings to `VideoPlayerSettingsModel` for `enableEdgeGestures` (default: true) and `reverseEdgeGestures` (default: false).

- Updated localization (`app_en.arb`) to include new strings and descriptions for edge gestures settings and brightness indicator.

- Consolidated the gestures into their separate subcategory called `Gestures` in the settings page as it was becoming cluttered.

**Volume Synchronization Improvements:**

- Enhanced `VideoPlayerSettingsProviderNotifier` to sync internal volume with the system volume on Android/iOS (using the `volume_controller` package), including initialization, event listening, and cleanup. This ensures volume changes via hardware buttons are reflected in the app and vice versa.

- Modified `setVolume` and `steppedVolume` methods to update both the internal model and the system volume on supported platforms.

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/discussions/292

## Checklist

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -61,7 +61,7 @@ PODS:
   - local_auth_darwin (0.0.1):
     - Flutter
     - FlutterMacOS
-  - mdk (0.35.0)
+  - mdk (0.35.1)
   - media_kit_libs_ios_video (1.0.4):
     - Flutter
   - media_kit_video (0.0.1):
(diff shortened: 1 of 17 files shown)


schembriaiden:
I have only tested this on Android as I have an Android phone. If somebody can test this on iOS that would be great.

For iOS you cannot test this on a simulator because the `volume_controller` package does not support the simulator:

https://pub.dev/packages/volume_controller

PartyDonut (approved):
Looks good to me, tested on iOS had to downgrade the volume_controller package version. Works/builds fine after that.

And changed the text color to always be "white" in-line with the other controls

PR #908, opened 2026-04-05
docs: add tested on platforms to PR template

schembriaiden:
## Pull Request Description

This pull request adds a new "Tested On" checklist section to the pull request template, making it easier for contributors to indicate which platforms they have tested their changes on.

Not sure if you would want it, but I think it would make it easier to indicate which platforms the author has tested on, and which might need to be tested.

## Issue Being Fixed

Resolves #issue-number

## Checklist


PartyDonut (approved):
Good idea 👍🏼

PR #915, opened 2026-04-06
fix: macOS shortcut handling

### LINKED ISSUE #882: ⌘+Q not closing the app on Mac
### Description

I guess this has to do with the app being cross-platform and some mac features not being supported, yet. ⌘+Q is currently not quitting the app but clicking on the red X does. At the same time closing Fladder always causes the app to freeze for some seconds.

### Reproduction steps

```bash
1. Open Fladder on Mac
2. ⌘+Q
```

### Screenshots

_No response_

### Logs

```bash

```

### Platform

macOS

### App Version

0.10.2

### Jellyfin server

10.11.6

schembriaiden:
## Pull Request Description

This pull request improves the flexibility and accuracy of keyboard shortcut handling by enhancing how modifier keys are compared and recognized. The main changes introduce a new method for matching modifier keys, expand support for "super" keys, and update the logic for detecting key combinations throughout the codebase.

Key combination logic improvements:

* Added the `modifierMatches` static method to `KeyCombination` to allow comparison of modifier keys by their functional group (shift, alt, ctrl, super), not just by exact key identity. This enables consistent recognition of different physical keys that serve the same modifier function (e.g., left/right shift, meta, super).
* Updated the `containsSameSet` method in `KeyCombination` to use `modifierMatches` for more robust comparison of key combinations.
* Expanded the set of recognized modifier keys by introducing `superKeys` (including meta and super keys) and incorporating them into the `modifierKeys` set.

Input handling updates:

* Modified `_InputHandlerState` in `input_handler.dart` to use `KeyCombination.modifierMatches` when checking if a pressed modifier key matches a configured modifier, ensuring consistent behavior with the new comparison logic.

Previous PR (https://github.com/DonutWare/Fladder/pull/901) did not fix the issue as when I tried the nightly build on my Macbook the fix did not work.

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/issues/882

## Checklist

### DIFF
--- a/lib/models/settings/key_combinations.dart
+++ b/lib/models/settings/key_combinations.dart
@@ -102,10 +102,8 @@ abstract class KeyCombination with _$KeyCombination {
     if (pressedModifier == configuredModifier) return true;
     if (pressedModifier == null || configuredModifier == null) return false;
 
-    return (shiftKeys.contains(pressedModifier) && shiftKeys.contains(configuredModifier)) ||
-        (altKeys.contains(pressedModifier) && altKeys.contains(configuredModifier)) ||
-        (ctrlKeys.contains(pressedModifier) && ctrlKeys.contains(configuredModifier)) ||
-        (superKeys.contains(pressedModifier) && superKeys.contains(configuredModifier));
+    // Keep left/right distinction for Shift/Alt/Ctrl, but normalize macOS Command/Super variants.
+    return superKeys.contains(pressedModifier) && superKeys.contains(configuredModifier);
   }
 }
 
--- a/lib/models/settings/key_combinations.dart
+++ b/lib/models/settings/key_combinations.dart
@@ -31,7 +31,8 @@ abstract class KeyCombination with _$KeyCombination {
   }
 
   bool containsSameSet(KeyCombination other) {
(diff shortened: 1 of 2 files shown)


PartyDonut (changes requested):
Very strange, I just tested it and for some reason the "super/command" + q combination does work on my laptop keyboard but not on my external keyboard. Despite being the same keys on the same side. 

That being said, the changes look ok. But lets keep the shift/ctrl/alt left and right modifier keys separate. I think we should only combine the super/meta left and right keys.

schembriaiden:
> Very strange, I just tested it and for some reason the "super/command" + q combination does work on my laptop keyboard but not on my external keyboard. Despite being the same keys.

I have no idea what might be causing it. I have an M1 Macbook Air. Might be some differences in the OS/drivers between variants I guess. 

> That being said, the changes look ok. But lets keep the shift/ctrl/alt left and right modifier keys separate. I think we should only combine the super/meta left and right keys.

Should be fixed.

PartyDonut (approved):
Thanks for the lightning quick fix!

> I have no idea what might be causing it. I have an M1 Macbook Air. Might be some differences in the OS/drivers 
> between variants I guess.

M2 air here so should not be that different. I do have the international keyboard version not sure if that changes anything.

PR #922, opened 2026-04-07
fix: prevent navigation and focus issues by checking route state

### LINKED ISSUE #694: 🐛  Hitting escape when any of the player menus(subtitles,audios, transcoding etc.) is open causes the player to crash
### Description

In the web ui when the player is open and a player menu is open like the audio track etc. when you hit escape the video vanishes and you're left with a black player with 0:00 on timestamp, nothing else works after this. Normal escape behaviour when menus aren't open is the same as stop button(it closes the player completely) however when a menu is open this behaviour is buggy.

### Reproduction steps

```bash
1. Go to Web UI 
2. Click on any item to play it
3. Hit 3 dots, or audio track or subtitle selection icons to open up any of these menus
4. Hit the escape key 
5. The player is completely black and not recoverable, you should close player altogether and open the movie/series again.
6. The expected behaviour should be for the menu to close upon hitting escape same as clicking outsite the menu
```

### Screenshots

Image

### Logs

```bash

```

### Platform

Web

### App Version

_No response_

### Jellyfin server

10.10.7

schembriaiden:
## Pull Request Description

This pull request improves the handling of focus and exit actions in the video player and input handler components by ensuring that certain actions only occur when the current route is active. This helps prevent unintended behavior when the widget is not the topmost route.

**Route-awareness improvements:**

* In both `_TvPlayerControlsState` (`tv_player_controls.dart`) and `_DesktopControlsState` (`video_player_controls.dart`), the exit hotkey now checks if the current route is active before closing the player, preventing accidental closure when not on the main route.
* In `_InputHandlerState` (`input_handler.dart`), autofocus is now only triggered if the current route is active, avoiding unwanted focus changes when the widget is not visible.

As an example, when going into one of the sub-menus while in the player, such as `Subtitles` or `Audio`, you can press `Escape` to go back to the player without actually exiting the player. If you are in the player itself it will then stop the player.

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/issues/694

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [x] Linux
- [ ] Windows
- [ ] macOS
- [x] Web

## Checklist

### DIFF
--- a/lib/screens/video_player/tv_player_controls.dart
+++ b/lib/screens/video_player/tv_player_controls.dart
@@ -710,8 +710,12 @@ class _TvPlayerControlsState extends ConsumerState<TvPlayerControls> {
         }
         return true;
       case VideoHotKeys.exit:
-        closePlayer();
+        if (ModalRoute.of(context)?.isCurrent == true) {
+          closePlayer();
+          return true;
+        }
         return false;
+
       case VideoHotKeys.mute:
         if (volume != 0) {
           previousVolume = volume;
--- a/lib/screens/video_player/video_player_controls.dart
+++ b/lib/screens/video_player/video_player_controls.dart
@@ -929,8 +929,12 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
         }
         return true;
       case VideoHotKeys.exit:
-        closePlayer();
+        if (ModalRoute.of(context)?.isCurrent == true) {
+          closePlayer();
+          return true;
+        }
         return false;
+
       case VideoHotKeys.mute:
         if (volume != 0) {
           previousVolume = volume;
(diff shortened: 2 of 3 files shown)


schembriaiden:
@PartyDonut seeing this issue https://github.com/DonutWare/Fladder/issues/930, do you think this would also make sense as a change in the player?

If the player is fullscreen, when the user first hits Escape it takes them out of fullscreen, and if they hit Escape again it stops the player.

Currently when a player hits Escape once while fullscreen it goes out of fullscreen and stops the player, so I thought adding the above change would help.

PartyDonut (approved):

PartyDonut:
> @PartyDonut seeing this issue #930, do you think this would also make sense as a change in the player?
> 
> If the player is fullscreen, when the user first hits Escape it takes them out of fullscreen, and if they hit Escape again it stops the player.

This is the other way around right? Or at least for me it closes the player first and hitting escape again exits full-screen (on macOS). This might actually depend on this system in which case the default "escape" shortcut should probably be changed?

How it works now from the PR is probably better close the player first and then let the system take over exiting full-screen.

PR #928, opened 2026-04-09
feat: Add hotkeys to step through frames

peanut250:
## Pull Request Description
Adds hotkeys to step through individual frames similarly to youtube, mpv and other players (, to step backward and . to step forward).

## Issue Being Fixed

Addresses https://github.com/DonutWare/Fladder/discussions/892

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist


lib/models/settings/video_player_settings.dart
```diff
@@ -18,6 +18,8 @@ enum VideoHotKeys {
   playPause,
   seekForward,
   seekBack,
+  stepForward,
+  stepBack,
```
PartyDonut, line 22: When a new enumerator value is added you need to run the flutter build_runner. Currently the hotkeys don't save because of a null error because the enum does not exist when serializing.

lib/screens/video_player/video_player_controls.dart
```diff
@@ -949,6 +968,18 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
       case VideoHotKeys.prevChapter:
         ref.read(videoPlayerSettingsProvider.notifier).prevChapter();
         return true;
+      case VideoHotKeys.stepForward:
+        if (playing) {
+          return false;
```
PartyDonut: Not sure about this logic? I get not wanting to skip frames during playback. But we should either show the user a message to pause first.

Should the shortcut pause the application or show a message to pause first?

My preference would be pause the player and skip showing a small notification of what just happened similar to the other video pop-ups (volume etc.). And if the video was playing resume after a few seconds.

lib/screens/video_player/video_player_controls.dart
```diff
@@ -681,6 +681,24 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
...
+
+  void stepForward(WidgetRef ref) {
+    final mediaPlayback = ref.read(mediaPlaybackProvider);
+    final framerate = ref.read(playBackModel.select((value) => value?.mediaStreams?.videoStreams.first.frameRate));
+    final step = (1000000.0 / framerate!).round();
+    resetTimer();
+    final newPosition = (mediaPlayback.position.inMicroseconds + step).clamp(0, mediaPlayback.duration.inMicroseconds);
+    ref.read(videoPlayerProvider).seek(Duration(microseconds: newPosition));
+  }
+
```
PartyDonut, line 703: This works, but my preference would be to implement the "steps" in both libmpv and libmdk if possible.

lib/screens/video_player/video_player_controls.dart
```diff
@@ -681,6 +681,24 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
...
+    final step = (1000000.0 / framerate!).round();
+    resetTimer();
+    final newPosition = (mediaPlayback.position.inMicroseconds - step).clamp(0, mediaPlayback.duration.inMicroseconds);
+    ref.read(videoPlayerProvider).seek(Duration(microseconds: newPosition));
+  }
+
+  void stepForward(WidgetRef ref) {
+    final mediaPlayback = ref.read(mediaPlaybackProvider);
+    final framerate = ref.read(playBackModel.select((value) => value?.mediaStreams?.videoStreams.first.frameRate));
+    final step = (1000000.0 / framerate!).round();
```
PartyDonut: Not a big fan of bang operators. Lets use a guard statement and return early.

PartyDonut (changes requested):
Thanks for implementing this, added a few comments.

lib/screens/video_player/video_player_controls.dart
peanut250: Not really any use case I can think of for it to work during playback. This behavior is the same as the youtube player. IMO a bit overkill to have a unique pop up and probably undesirable to auto resume, I can just have it pause the video like mpv.

PartyDonut: Sounds good lets keep it similar to how mpv does things.

lib/screens/video_player/video_player_controls.dart
peanut250: Thought about it, I don't see the reason to distinguish between step and seek in libmpv/libmdk since the seek functions expect the calculated seek position as a parameter. "step" is just a calculated seek position. IMO a bit cleaner for step to use these seek functions and reduces the maintenance burden if any other backends are supported.

PartyDonut: Sorry I should probably have been clearer about this.

We should at least move this functionality to the `base_player.dart` abstract class. That way you can have the same functionality for all of them. But implement a specific behavior for any player that needs it.

peanut250: I'm not sure about that either, the step seek position needs to be precalculated to update the UI in `media_control_wrapper`, similar to normal seek. There shouldn't be a need for any player to implement specific behavior just for step since both step and seek are covered by `seek(position)`. Maybe the logic can be moved to `media_control_wrapper`?

PartyDonut: There could be a instance of the player support seeking frame by frame so we could call that instead of doing the calculation ourselves. That being said this improvement is not a blocking issue, mostly a preference.

Fine to leave it as is for now.

PartyDonut (approved):
Looks good to me, thanks for implementing the feedback

PR #931, opened 2026-04-10
chore: split Android builds

schembriaiden:
## Pull Request Description

This pull request updates the Android build process to generate and handle split APKs for different CPU architectures, instead of a single universal APK. It also adds packaging exclusions for certain resources in the Gradle configuration. These changes optimize the app's distribution by reducing APK sizes and ensuring only necessary files are included.

**Android build and artifact handling improvements:**

* The build process now uses the `--split-per-abi` flag in the `flutter build apk` command, generating separate APKs for each supported ABI instead of a single universal APK.
* The workflow moves and renames each split APK (`armeabi-v7a`, `arm64-v8a`, and `x86_64`) individually, ensuring all variants are available as artifacts.
* The release artifact list is updated to include all split APKs, instead of just one universal APK.

**Gradle configuration enhancements:**

* The `packagingOptions` block is added to the `release` build type in `android/app/build.gradle` to exclude `assets/dexopt/baseline.prof` and `assets/dexopt/baseline.profm` from the packaged APKs, reducing unnecessary resources in the final builds.

All of these should allow Reproducible Builds on F-Droid.

## Issue Being Fixed

Relates to: https://github.com/DonutWare/Fladder/discussions/553#discussioncomment-16515800

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -171,26 +171,33 @@ jobs:
       - name: Build Android APK (PRs)
         if: needs.fetch-info.outputs.build_type == 'development'
         run: |
-          flutter build apk --debug --build-number=${{github.run_number}} --flavor production
+          flutter build apk --debug --split-per-abi --build-number=${{github.run_number}} --flavor production
 
       - name: Build Android APK and AAB
         if: needs.fetch-info.outputs.build_type != 'development'
         run: |
-          flutter build apk --release --build-number=${{github.run_number}} --flavor production
+          flutter build apk --release --split-per-abi --build-number=${{github.run_number}} --flavor production
           flutter build appbundle --release --build-number=${{github.run_number}} --flavor production
 
       - name: Rename APK for PR
         if: needs.fetch-info.outputs.build_type == 'development'
         run: |
           mkdir -p build/app/outputs/android_artifacts
-          mv build/app/outputs/flutter-apk/app-production-debug.apk "build/app/outputs/android_artifacts/${{ env.ARTIFACT_SUFFIX }}.apk"
+          # Move split debug APKs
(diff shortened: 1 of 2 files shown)


PartyDonut:
Not sure, this adds more to maintain on what is already a pretty big and complicated workflow. 

Also for f-droid the type of release doesn't matter right? The only thing they want is that Fladder builds the exact same on here as it does on their build. So only the mdk library would be the biggest hurdle.

schembriaiden:
While I was implementing it I noticed that this adds more complexity, I had just opened this just in case you would want to do it.

They had told me to try and mirror this repo which is currently on f-droid and has reproducible builds: https://github.com/Predidit/Kazumi/

This is the F-Droid metadata for Kazumi: https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/com.predidit.kazumi.yml

And this is the one I was trying to implement: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/34689/diffs

Having [reproducible builds](https://f-droid.org/docs/Reproducible_Builds/) is not a requirement, its just a suggestion but I think it would ensure that any build made on GitHub and made on F-Droid would be exactly the same. I don't know what issues could arise from not having reproducible builds but I was told that if we decide not to go with it now, we cannot add it in later.

PartyDonut (approved):

PartyDonut:
Probably better to keep it similar to a known published repo where possible. 

Do we still need to separate the mdk libraries build process?

schembriaiden:
> Do we still need to separate the mdk libraries build process?

I am not 100% sure I will check it out.

PR #942, opened 2026-04-12
fix: improve subpath handling in docker

schembriaiden:
## Pull Request Description

This pull request improves the Docker setup and entrypoint logic for the Fladder web application, focusing on better handling of custom web paths and permissions. The main changes involve normalizing and applying the `FLADDER_WEBPATH` variable more robustly, dynamically generating the Nginx configuration based on the deployment context, and ensuring correct file ownership for Nginx.

**Entrypoint and Nginx configuration improvements:**

* The `docker-entrypoint.sh` script now normalizes the `FLADDER_WEBPATH` variable, ensuring it always starts and ends with a single slash, and updates the `<base href>` in `index.html` accordingly.
* The script dynamically generates the Nginx configuration in `/etc/nginx/conf.d/default.conf` depending on whether Fladder is served at the root or a subpath, and selects the correct port based on the user (root or rootless).

**File and directory permissions:**

* Both `Dockerfile` and `Dockerfile-rootless` now ensure that `/usr/share/nginx/html` and `/etc/nginx/conf.d` are owned by the `nginx` user, preventing permission issues at runtime.
* Ownership of the `docker-entrypoint.sh` script is explicitly set to `nginx:nginx` in the rootless Dockerfile for consistency.

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/discussions/917#discussioncomment-16532230

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [x] Web

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,8 @@ COPY build/web /usr/share/nginx/html
 COPY docker-entrypoint.sh /docker-entrypoint.sh
 
 RUN mkdir -p /usr/share/nginx/html/assets/config && \
-    chmod +x /docker-entrypoint.sh
+    chown -R nginx:nginx /usr/share/nginx/html && \
+    chmod +x /docker-entrypoint.sh && \
+    chown -R nginx:nginx /etc/nginx/conf.d
 
 CMD ["/docker-entrypoint.sh"]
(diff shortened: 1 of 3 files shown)


Zepmann:
Consider an alternative: do not move files. Adjust the nginx configuration instead. That is more in line with other configuration files adjusted when starting a container.

Zepmann:
@schembriaiden 

Tested from my end. Looks good! Configured the environment variable:

```
FLADDER_WEBPATH=/fladder/
```

Without any workarounds from a reverse proxy running in front, Fladder is successfully served from `https://example.com/fladder/` (also works without a backslash at the end).

PartyDonut (approved):
Looks good to me thanks for the quick implementation.

And thanks @Zepmann voor testing it out 👍🏼

PR #943, opened 2026-04-12
chore: simplify metadata pipeline

schembriaiden:
## Pull Request Description

This PR aims to simplify the pipeline as after the last release failed I noticed a flaw with it.

The actual commits of the metadata would not be in the release itself which would be pointless when updatng Flatpak and F-Droid later, so this PR should solve that by updating the metadata before the actual build is run.

## Issue Being Fixed

Fixes [GitHub Action](https://github.com/DonutWare/Fladder/actions/runs/24310174700) release issue

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,94 @@
+name: Prepare Release
+
+on:
+  workflow_dispatch:
+
+jobs:
+  prepare-release:
+    name: Update Metadata and Tag
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate Bot Token
+        id: generate-token
+        uses: actions/create-github-app-token@v3.1.1
+        with:
+          app-id: ${{ secrets.FLADDER_BOT_APP_ID }}
+          private-key: ${{ secrets.FLADDER_BOT_APP_PRIVATE_KEY }}
+
+      - name: Checkout repository
+        uses: actions/checkout@v4.1.1
+        with:
+          ref: develop
+          fetch-depth: 0
+          token: ${{ steps.generate-token.outputs.token }}
+
+      - name: Extract Version
+        id: version
+        run: |
+          # Extract version from pubspec.yaml (e.g., 0.10.3+1)
+          FULL_VERSION=$(grep '^version: ' pubspec.yaml | sed 's/version: //')
+          # Remove build number (e.g., 0.10.3)
+          VERSION_NAME=$(echo "$FULL_VERSION" | cut -d'+' -f1)
+          # Exact Build Number (e.g., 1)
+          BUILD_NUMBER=$(echo "$FULL_VERSION" | cut -d'+' -f2)
+          
+          echo "version=${VERSION_NAME}" >> "$GITHUB_OUTPUT"
+          echo "build_number=${BUILD_NUMBER}" >> "$GITHUB_OUTPUT"
+          echo "Version: $VERSION_NAME (Build: $BUILD_NUMBER)"
+          
+          # Check if this version is already tagged
+          if git rev-parse "v${VERSION_NAME}" >/dev/null 2>&1; then
+            echo "is_new_version=false" >> "$GITHUB_OUTPUT"
+            echo "Status: Version is already tagged."
+          else
+            echo "is_new_version=true" >> "$GITHUB_OUTPUT"
+            echo "Status: New version detected."
+          fi
+
+      - name: Update metainfo.xml
+        if: steps.version.outputs.is_new_version == 'true'
+        env:
+          CHANGELOG_MSG: "See the full changelog at: https://github.com/${{ github.repository }}/releases/tag/v${{ steps.version.outputs.version }}"
+        run: |
+          VERSION="${{ steps.version.outputs.version }}"
+          DATE=$(date +%Y-%m-%d)
+          NEW_RELEASE="  <releases>\n    <release version=\"${VERSION}\" date=\"${DATE}\">\n      <description>\n        <p>${{ env.CHANGELOG_MSG }}</p>\n      </description>\n    </release>\n  </releases>"
+          perl -i -0777 -pe "s/  <releases>.*?<\/releases>/$NEW_RELEASE/s" flatpak/nl.jknaapen.fladder.metainfo.xml
+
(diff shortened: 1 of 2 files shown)


PartyDonut:
Thanks for the quick update. But I think it might be better to just create a pull-request, that way we don't have to create a new token. And we have the ability to view and/or edit the changes if need be.

schembriaiden:
> Thanks for the quick update. But I think it might be better to just create a pull-request, that way we don't have to create a new token. And we have the ability to view and/or edit the changes if need be.

The only reason I worry about this is due to the fact that as I mentioned F-Droid and Flatpak build from source, and if we do go ahead with the PR the release commit would not include the metadata updates which would make them a bit redundant.


While it would be possible to then move the commit number, F-Droid would probably have issues because the Reproducible builds would likely fail then due to not having the same commit.

EDIT: The last force-push was just a commit to simplify the COMMIT_MSG so that its the same for both fastlane and metainfo, nothing to do with the above.

What if we force-push to the tag only, pull the tag changes and create the release from the tag, and then create a PR to merge to develop? Like that we should not need to create another token (This probably would not work if you have a tag ruleset as well)

PartyDonut:
Yeah makes sense we need to keep it in sync. I've created a github app that has the needed permissions that should work as well right?

Would be better then creating specific tokens. And probably a bot we could re-use for other things.

 app-id: ${{ secrets.FLADDER_BOT_APP_ID }}
 private-key: ${{ secrets.FLADDER_BOT_APP_PRIVATE_KEY }}

schembriaiden:
> Yeah makes sense we need to keep it in sync. I've created a github app that has the needed permissions that should work as well right?

Yep this should probably work, as long as the app is added to the "Allow bypass" list in the develop branch protection settings, or that the App has sufficient permissions to push to protected branches.

I have updated it to use the App instead

PartyDonut:
Thanks again for the quick implementation. However I'm still not quite sure this is the way to go.

Instead of re-writing the commit history it would probably be better to prepare the meta-info first.
Maybe we should have a prepare release workflow. 

This fetches the version from the pubspec creates the meta-info files. 

When it's done push the changes and the newly created tag to develop, and the release workflow then continues like normal.

schembriaiden:
> Thanks again for the quick implementation. However I'm still not quite sure this is the way to go.
> 
> Instead of re-writing the commit history it would probably be better to prepare the meta-info first. Maybe we should have a prepare release workflow.
> 
> This fetches the version from the pubspec creates the meta-info files.
> 
> When it's done push the changes and the newly created tag to develop, and the release workflow then continues like normal.


Just to confirm, would a workflow like this suffice? If this sounds good to you I will push the changes.

1.  Update the version number in `pubspec.yaml` and push the change to the `develop` branch.
2.  New workflow called `prepare-release` runs which:
    - Fetches the new version.
    - Updates `metainfo.xml` and Fastlane changelogs.
    - Commits those changes back to `develop`.
    - Creates and pushes the release tag (e.g., `v0.10.3`).
5.  The main `build` workflow will detect the new tag pushed by the bot and immediately start the release process.

PartyDonut:
> > Thanks again for the quick implementation. However I'm still not quite sure this is the way to go.
> > Instead of re-writing the commit history it would probably be better to prepare the meta-info first. Maybe we should have a prepare release workflow.
> > This fetches the version from the pubspec creates the meta-info files.
> > When it's done push the changes and the newly created tag to develop, and the release workflow then continues like normal.
> 
> Just to confirm, would a workflow like this suffice? If this sounds good to you I will push the changes.
> 
>     1. Update the version number in `pubspec.yaml` and push the change to the `develop` branch.
> 
>     2. New workflow called `prepare-release` runs which:
>        
>        * Fetches the new version.
>        * Updates `metainfo.xml` and Fastlane changelogs.
>        * Commits those changes back to `develop`.
>        * Creates and pushes the release tag (e.g., `v0.10.3`).
> 
>     3. The main `build` workflow will detect the new tag pushed by the bot and immediately start the release process.

Sounds good for the most part, but lets skip step 1. Would be better to update the version number manually. 

Then in actions we start the workflow prepare-release.

schembriaiden:
> Sounds good for the most part, but lets skip step 1. Would be better to update the version number manually.

What I meant here was you update the version yourself in `pubspec.yaml` yourself and push this to develop, then this triggers the workflow run in step 2, which I think aligns with what you're saying

schembriaiden:
Should be done, let me know if its correct now.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,92 @@
+name: Prepare Release
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - develop
+    paths:
+      - pubspec.yaml
```
PartyDonut: Let's remove this.as in this case adding new dependencies would also trigger this workflow.

Best to keep it as a workflow_dispatch only.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,92 @@
...
+          git add fastlane/metadata/android/en-US/changelogs/
+          
+          if git diff --staged --quiet; then
+            echo "No changes to metadata"
+          else
+            git commit -m "chore: prepare release $VERSION"
+            git push origin develop
+          fi
+          
+          # Push the tag
```
PartyDonut: For consistency lets first merge develop into main.
Then create a release tag, this keeps the release pattern in-line with existing release logic.

PartyDonut (changes requested):

.github/workflows/prepare-release.yml
schembriaiden: Should be done, hopefully I did it correctly

PartyDonut (approved):
Looks good to me thanks for making all the changes 👍🏼

schembriaiden:
I saw you ran the PR change line 56 to this:

`perl -i -0777 -pe "s@  <releases>.*?</releases>@$NEW_RELEASE@s" flatpak/nl.jknaapen.fladder.metainfo.xml`

PR #944, opened 2026-04-12
chore: remove metainfo download

schembriaiden:
## Pull Request Description

Apologies for all the leftover stuff, I think this should be the last remaining thing now.

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/actions/runs/24315476768/job/70992602970

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -401,13 +401,6 @@ jobs:
         with:
           submodules: true
 
-      - name: Download updated metainfo
-        if: needs.fetch-info.outputs.build_type == 'release'
-        uses: actions/download-artifact@v4
-        with:
-          name: fladder-metainfo
-          path: flatpak
-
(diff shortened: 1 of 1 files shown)


PartyDonut:
Thanks for the PR, already removed it 👍🏼 though. I also changed the fastlane changelog to use the version number instead of build number.

Because with the prepare-release workflow that number will always be 1. Normally the Github runner number is used there, which is especially nice when using the nightly builds.

schembriaiden:
> Because with the prepare-release workflow that number will always be 1. Normally the Github runner number is used there, which is especially nice when using the nightly builds.

Oh you're right, thanks :)

PR #948, opened 2026-04-14
feat(library): add persistent recursive toggle setting with session m…

sodacoco91:
…emory

- Add recursiveByDefault setting (default: false) in ClientSettingsModel
- Add sessionRecursiveOverride to track user's manual toggle during session
- Modify LibrarySearchNotifier to respect session state over default setting
- Change navigateTo() to pass recursive: null instead of hardcoded true
- Add UI toggle in Settings -> Client Settings -> Advanced
- Add localization strings for the new setting

Behavior:
- App startup: uses recursiveByDefault value
- Manual toggle: sessionRecursiveOverride is set, persists across all libraries
- Next app restart: session state cleared, falls back to recursiveByDefault

## Issue Being Fixed

Resolves #issue-number

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

### DIFF
--- a/.github/workflows/Build-Fladder-Windows-(Manual).yml
+++ b/.github/workflows/Build-Fladder-Windows-(Manual).yml
@@ -1,0 +1,82 @@
+name: Build Fladder Windows (Manual)
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: '要编译的分支'
+        required: false
+        default: 'custom-dev'
+        type: string
+
+concurrency:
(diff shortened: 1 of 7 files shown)


PartyDonut:
Hi thanks for opening a PR but this is not something we should handle this way.
The better solution would be more sensible defaults for different library types. 

Your PR also contains files that where probably meant for a other commit?

PR #845, opened 2026-03-15
feat: Dynamically update desktop window title

schembriaiden:
## Pull Request Description

This pull request introduces a new context-aware window title system for the application, ensuring that the window title reflects the current navigation state and playback context. The changes span across model classes, UI scaffolds, and platform-specific code, enhancing user experience and accessibility.

## Issue Being Fixed

Resolves #issue-number

## Checklist

- [ ] If a new package was added, did you ensure it works for all supported platforms? Is the package well maintained
- [ ] Check that any changes are related to the issue at hand.

PR #949, opened 2026-04-15
fix: screen flicker when stopping floating player on small devices

schembriaiden:
## Pull Request Description

This pull request updates how the application window title is set in the `lib/main.dart` file. Instead of using the `onGenerateTitle` property of `MaterialApp.router`, the window title is now managed by wrapping the app's content in a `Title` widget that dynamically responds to the `windowTitleProvider`. This change allows for more flexible and reactive window title updates.

**Window Title Management:**

* Replaced the use of `onGenerateTitle` in `MaterialApp.router` with a static `title` property, and moved dynamic title handling to a `Title` widget that listens to `windowTitleProvider` via a `Consumer`.
* Wrapped the main app content with the `Title` widget to ensure the window title updates reactively based on provider changes.

## Issue Being Fixed

I noticed an issue on Android where when pressing stop on the floating player the screen flickered. After checking the logs I found this:

```dart
2026-04-15 08:35:11.475 19271-19271 flutter                 nl.jknaapen.fladder.dev              I  SEVERE: 2026-04-15 10:35:11.475711: Flutter error: Looking up a deactivated widget's ancestor is unsafe.
At this point the state of the widget's element tree is no longer stable.
To safely refer to a widget's ancestor in its dispose() method, save a reference to the ancestor by calling dependOnInheritedWidgetOfExactType() in the widget's didChangeDependencies() method.
2026-04-15 08:35:11.476 19271-19271 flutter                 nl.jknaapen.fladder.dev              I  #0      Element._debugCheckStateIsActiveForAncestorLookup.<anonymous closure> (package:flutter/src/widgets/framework.dart:4954:9)
#1      Element._debugCheckStateIsActiveForAncestorLookup (package:flutter/src/widgets/framework.dart:4968:6)
#2      Element.dependOnInheritedWidgetOfExactType (package:flutter/src/widgets/framework.dart:4988:12)
#3      PopupMenuTheme.of (package:flutter/src/material/popup_menu_theme.dart:278:10)
#4      PopupMenuButtonState._positionBuilder (package:flutter/src/material/popup_menu.dart:1612:62)
#5      _PopupMenuRoute.buildPage.<anonymous closure> (package:flutter/src/material/popup_menu.dart:1054:32)
#6      _LayoutBuilderElement._rebuildWithConstraints.updateChildCallback (package:flutter/src/widgets/layout_builder.dart:233:74)
#7      BuildOwner.buildScope (package:flutter/src/widgets/framework.dart:3046:19)
#8      _LayoutBuilderElement._rebuildWithConstraints (package:flutter/src/widgets/layout_builder.dart:271:12)
#9      RenderAbstractLayoutBuilderMixin.layoutCallback (p
```

This indicated to me that it was probably caused by https://github.com/DonutWare/Fladder/pull/845, which has led to this change which resolves the issue.

Funnily enough this only happens on Android (at least from the platforms which I can test) and not on the desktops, I have no idea why.

## Tested On

- [x] Android
- [ ] Android TV
- [ ] iOS
- [x] Linux
- [ ] Windows
- [x] macOS
- [x] Web

Also tested with Russian translation as well and works fine.

## Checklist

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -15,6 +15,7 @@ import 'package:fladder/providers/crash_log_provider.dart';
 import 'package:fladder/providers/settings/client_settings_provider.dart';
 import 'package:fladder/providers/shared_provider.dart';
 import 'package:fladder/providers/sync_provider.dart';
+import 'package:fladder/providers/window_title_provider.dart';
 import 'package:fladder/routes/auto_router.dart';
 import 'package:fladder/theme.dart';
 import 'package:fladder/util/adaptive_layout/adaptive_layout.dart';
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -102,7 +102,7 @@ class _FladderApp extends ConsumerWidget {
           light: lightTheme,
           dark: darkTheme,
           child: MaterialApp.router(
-            onGenerateTitle: (context) => ref.watch(windowTitleProvider),
+            title: 'Fladder',
             theme: lightTheme,
             scrollBehavior: scrollBehaviour.copyWith(
               dragDevices: {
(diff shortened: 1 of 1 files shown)


PartyDonut (changes requested):
I'm not sure this is a fix we should use. While the Consumer does stop the entire application from rebuilding with the ref.watch.

Its still a ref.watch in a bit of a strange/problematic place. That's also the reason I removed it.

A better solution would probably be calling the `  SystemChrome.setApplicationSwitcherDescription(
      ApplicationSwitcherDescription(label: title),
    );` inside of the window_title_provider `update` function.

edit: Also not quite sure what the "screen flicker" issue is that the title mentions, and how it's related to the fix?

schembriaiden:
> edit: Also not quite sure what the "screen flicker" issue is that the title mentions, and how it's related to the fix?

The issue is, if you use the current version 0.10.3, when you minimise the player and you have the floating player, and you press the 3 dots and then press stop the screen flashes for a second. But it seems like the issue is now fixed in develop with your change.

I am not sure what exactly the `onGenerateTitle` used to do, was it used for the title bar at the top of the app which there is on some OS's?

Currently on develop it still auto generates window titles for me on Linux at least, so I'm not sure if the fix is needed anymore if the title thing isn't needed.

image

PartyDonut:
Let's close it for now. I don't see any changes with or without the onGenerateTitle on Android either. Thanks for having a look though.

PR #960, opened 2026-04-17
fix: video playback state synchronization

### LINKED ISSUE #787: 🐛 Playback seems to resume when screensave is active
### Description

When I pause playback for an extended period (around 20–30 minutes), the Android TV screensaver appears as expected. However, the video doesn’t seem to stay paused in the background. Instead, playback continues invisibly while the screensaver is active.

### Reproduction steps

```bash
1. Start watching something
2. Pause it, write down the progress
3. Wait until Android TV's screensaver is active (10 minutes?)
4. Wait a few more minutes
5. Resume playing and check the progress
```

### Screenshots

_No response_

### Logs

```bash

```

### Platform

Android TV

### App Version

_No response_

### Jellyfin server

10.11.6, linux mint

schembriaiden:
## Pull Request Description

This pull request improves the handling of playback state synchronization between the UI and the playback model in the video player feature. The main updates ensure that the playback state is accurately reflected and that the UI stays in sync with the underlying state.

**Playback state synchronization improvements:**

* Added a `_setupListeners` method in `_VideoPlayerState` to listen for changes to the `playing` state from `mediaPlaybackProvider` and update the local `playing` variable accordingly, ensuring the UI remains in sync with the playback state.
* Updated the call to `updatePlaybackPosition` in `VideoPlayerNotifier` to use the new `playing` value from the event, rather than the previous state, for more accurate state updates.

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/issues/787

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

### DIFF
--- a/lib/providers/video_player_provider.dart
+++ b/lib/providers/video_player_provider.dart
@@ -84,7 +84,7 @@ class VideoPlayerNotifier extends StateNotifier<MediaControlsWrapper> {
     mediaState.update(
       (state) => state.copyWith(playing: event),
     );
-    ref.read(playBackModel)?.updatePlaybackPosition(currentState.position, currentState.playing, ref);
+    ref.read(playBackModel)?.updatePlaybackPosition(currentState.position, event, ref);
   }
 
   Future<void> updatePosition(Duration event) async {
--- a/lib/screens/video_player/video_player.dart
+++ b/lib/screens/video_player/video_player.dart
@@ -59,10 +59,21 @@ class _VideoPlayerState extends ConsumerState<VideoPlayer> with WidgetsBindingOb
     super.dispose();
   }
 
+  void _setupListeners() {
+    ref.listenManual(
+      mediaPlaybackProvider.select((value) => value.playing),
+      (previous, next) {
+        playing = next;
+      },
+      fireImmediately: true,
+    );
+  }
+
   @override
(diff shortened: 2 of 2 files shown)


PartyDonut:
I'm not sure this is the fix for that users problem.
This would only add a small delay to the call itself I believe. 

The issue might actually be in the way playback is paused or at least how the signal is passed on to flutter.

schembriaiden:
> The issue might actually be in the way playback is paused or at least how the signal is passed on to flutter.

I think this should be the actual fix now, thanks for your feedback.

PartyDonut:
Sorry, I might have not been clear enough. When I said the issue is in how paused/playing is managed, I meant the playbackmodel helper.

The user seems to be reporting a mismatch between what the client visually shows and what is being send to jellyfin.
So the problem is likely inside of the playbackmodel.

Currently all players report back their current state inside of `video_player_provider.dart` this in turn tells the playbackmodel to report back to the server.

My best guess there is a race condition between the playbackmodel reporting the state (play/paused) and reporting the position.

In short: Fladder incorrectly sends "playing" to jellyfin, despite the player being paused.

PR #961, opened 2026-04-17
feat: add subtitle toggle hotkey

schembriaiden:
## Pull Request Description

This pull request introduces a new "toggle subtitles" hotkey feature to the video player, allowing users to quickly enable or disable subtitles using a keyboard shortcut. The implementation includes localization, hotkey mapping, and logic for toggling subtitles, including handling multiple subtitle tracks and remembering the user's last selection.

**New Subtitle Toggle Feature:**

* Added a new `toggleSubtitles` hotkey to the `VideoHotKeys` enum, mapped by default to the "T" key (`LogicalKeyboardKey.keyT`).
* Implemented the `_toggleSubtitles` method in `_DesktopControlsState` to handle toggling subtitles on/off, supporting both single and multiple subtitle tracks and remembering the last selected subtitle.
* Integrated the new hotkey into the video player controls so pressing the hotkey triggers subtitle toggling.

**Localization:**

* Added the `toggleSubtitles` string to the English localization file (`app_en.arb`) for display in the UI and settings.
* Updated the `VideoHotKeys` label mapping to include the new `toggleSubtitles` key for proper display in settings menus.

**Codebase Maintenance:**

* Added the necessary import for `media_streams_model.dart` to support subtitle toggling logic.
* Introduced a `_lastSelectedSubtitleIndex` state variable to remember the user's last subtitle choice across toggles.

Logic is as follow:

Press toggle:

1. If there is only 1 type of subtitle it just uses it
2. If there is multiple subtitle types:
	1. First time it asks you to select the one you want
	2. After it keeps using the same one.

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/discussions/957

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist


lib/screens/video_player/video_player_controls.dart
```diff
@@ -889,6 +893,58 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
     _vDragLastValue = null;
   }
 
+  void _toggleSubtitles() {
```
PartyDonut, line 916: I'm not entirely sure about the logic here. Does it make sense for the user to press the "toggle" button to all of a sudden ask what subtitles they want to toggle in-between?

Would it make more sense to just do nothing if the subtitles are already disabled and only allow toggling when a subtitle is enabled?

No strong opinion on my end just thinking out loud.

lib/models/settings/video_player_settings.dart
```diff
@@ -241,6 +243,7 @@ Map<VideoHotKeys, KeyCombination> get _defaultVideoHotKeys => {
           VideoHotKeys.takeScreenshot => KeyCombination(key: LogicalKeyboardKey.keyG),
           VideoHotKeys.takeScreenshotClean =>
             KeyCombination(key: LogicalKeyboardKey.keyG, modifier: LogicalKeyboardKey.controlLeft),
+          VideoHotKeys.toggleSubtitles => KeyCombination(key: LogicalKeyboardKey.keyT),
```
PartyDonut, line 268: Silly question but why `T`? Youtube for instance uses the `C` button for "Captions"

schembriaiden: I was initially going to go with `C` to mirror Youtube's logic, but since the request suggested `T` I just went with that, we can do whatever I don't mind.

lib/screens/video_player/video_player_controls.dart
schembriaiden: I am not against the idea, but I think this might lead users to open issues on GitHub saying that the toggle key is not working, when in fact it would be working. Then from experience I know that most people don't check previous issues before opening one which would lead to a bunch of duplicate issues regarding this.

PartyDonut (changes requested):
It seems to not work entirely like expected.

When I have a subtitle selected it still opens the subselection.
After then selecting that subtitle it toggles on/off fine. 

When off is the initial selection and I press the toggle it shows subselection. After selecting something toggling works like expected.

But after then selecting a different subtitle toggling still toggles between the initial selection and off. Not the new selection and off.

schembriaiden:
All the cases you mentioned should be fixed now.

PartyDonut (approved):
Thanks for fixing the issues. Seems to work as intended now.

PR #963, opened 2026-04-17
feat: Chapter image trickplay/timeline fallback

matt-hb:
## Pull Request Description

First actual feature PR, chapter images have been bothering me a bunch so I'm implementing a trickplay fallback as per https://github.com/DonutWare/Fladder/discussions/875 .
Would love if someone could try it out on bigger screens on Android TV to see how it looks, since currently I am blurring trickplay images a bit as they are lower resolution (I think a bit of blur is nicer than pixelation).
Also added caching for trickplay images, because we're fetching them a bunch.

## Issue Being Fixed

Placeholder when there are no chapter images available, either via fallback to trickplay images or a timeline if trickplay images are not available either.

## Screenshots / Recordings

Before (yucky placeholders):
image

After (with trickplay fallback):
image

After (with timeline):
TODO

## Tested On

- [x] Android
- [ ] Android TV
- [x] Desktop
- [x] Web

## Checklist

- [ ] Create timeline widget
- [ ] Implement in VideoPlayerChapters?


matt-hb:
looks like some old commits from another PR are still there, they don't do anything for the diff though, my bad!

PartyDonut:
Just did a test for Android TV it's quite blurry. Does not look great to be honest.

Another solution would be to change the filterquality inside of the `_TrickplayPainter` from `trick_play_image.dart`

image

matt-hb:
> Another solution would be to change the filterquality inside of the `_TrickplayPainter` from `trick_play_image.dart`

Good call, I misunderstood low level drawing and size inheriting before, but after checking this is in fact a lot better.

PR #964, opened 2026-04-18
fix: audio glitches on Android

### LINKED ISSUE #762: 🐛 Audio glitching with mpv player on Android
### Description

When using the mpv player on Android (regardless of in-app settings I tried), there are audio glitching issues that happen every 10-30 seconds. It's never at the same section of audio, so for example if you hear an audio glitch then go back 10 seconds you're not going to hear it glitch at the same spot. When using the native player there are no issues.

I researched this a little bit and it seems like it's either because of audio buffer underrun or a clock synchronization drift.

### Reproduction steps

```bash
1. Set the video player to mpv
2. Start watching a movie
3. Hear audio glitches (doesn't matter if it's direct or transcoded audio)
4. Set the video player to native
5. See that the audio glitches are removed
```

### Screenshots

_No response_

### Logs

```bash

```

### Platform

Android

### App Version

v0.9.0

### Jellyfin server

10.11.6

schembriaiden:
## Pull Request Description

This pull request enhances the audio output stability on Android devices by configuring the underlying MPV player to use the `audiotrack` audio output when running on Android. This should help address audio playback glitches that have been reported.

Platform-specific configuration:

* In `lib/wrappers/players/lib_mpv.dart`, when initializing the MPV player on Android (`TargetPlatform.android`), the code now sets the `ao` (audio output) property to `audiotrack` for improved stability.

I would like your opinion for this "fix" if this is something you would want or not? But according to the user it does work.

## Issue Being Fixed

Resolves https://github.com/DonutWare/Fladder/issues/762#issuecomment-4273241812

## Tested On

- [x] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

### DIFF
--- a/lib/wrappers/players/lib_mpv.dart
+++ b/lib/wrappers/players/lib_mpv.dart
@@ -70,10 +70,13 @@ class LibMPV extends BasePlayer {
     }
 
     if (_player?.platform is mpv.NativePlayer) {
-      await (_player?.platform as dynamic).setProperty(
-        'force-seekable',
-        'yes',
-      );
+      final nativePlayer = _player!.platform as dynamic;
+      await nativePlayer.setProperty('force-seekable', 'yes');
+
+      if (defaultTargetPlatform == TargetPlatform.android) {
+        // Use audiotrack as it is generally more stable on modern Android
+        await nativePlayer.setProperty('ao', 'audiotrack');
+      }
     }
   }
 


lib/wrappers/players/lib_mpv.dart
```diff
@@ -70,10 +70,13 @@ class LibMPV extends BasePlayer {
...
-      await (_player?.platform as dynamic).setProperty(
-        'force-seekable',
-        'yes',
-      );
+      final nativePlayer = _player!.platform as dynamic;
+      await nativePlayer.setProperty('force-seekable', 'yes');
+
+      if (defaultTargetPlatform == TargetPlatform.android) {
+        // Use audiotrack as it is generally more stable on modern Android
+        await nativePlayer.setProperty('ao', 'audiotrack');
```
PartyDonut, line 78: I might be missing something but I don't see `audiotrack` in the mpv player audio output options?
https://mpv.io/manual/stable/#audio-output-drivers

schembriaiden: If you look at this it is there, it just doesn't seem to be in the docs: https://github.com/mpv-player/mpv/blob/master/meson.options

Maybe because its an Android-only feature?

PartyDonut: Yeah probably a Android only feature then. Works fine though was just curious.

PartyDonut (approved):

PR #965, opened 2026-04-18
fix: make pinch-to-zoom work again after adding edge gestures

schembriaiden:
## Pull Request Description

This pull request updates the touch gesture handling for vertical drag actions in the video player controls. The main improvement is to restrict vertical drag gestures for brightness and volume to the left and right edges of the screen, reducing accidental activations and aligning with common video player UX patterns.

**Touch gesture improvements:**

* Vertical drag gestures for brightness and volume are now only detected in dedicated left and right edge zones, each occupying 15% of the screen width and 80% of the height, instead of anywhere on the screen. This change helps avoid accidental triggers and provides a more intuitive user experience. (`lib/screens/video_player/video_player_controls.dart`)
* The `_handleVerticalDragStart` method now takes an explicit `isLeft` parameter, determined by which edge zone the gesture started in, rather than calculating the side based on the touch position. This makes the logic clearer and more maintainable. (`lib/screens/video_player/video_player_controls.dart`)

## Issue Being Fixed

After the last update I noticed an issue where the pinch-to-zoom was a bit inconsistent due to the new edge gestures I added, this update aims to fix that by making the edge gestures take up only the 15% edges of the screen and the middle is saved for the pinch to zoom. Before it was a 50/50 split for the edge gestures, which was a bit shortsighted from my end.

## Tested On

- [x] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

### DIFF
--- a/lib/screens/video_player/video_player_controls.dart
+++ b/lib/screens/video_player/video_player_controls.dart
@@ -135,11 +135,24 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
                         : _handleDoubleTapSeek,
                     onLongPressStart: initInputDevice == InputDevice.touch ? _handleLongPressStart : null,
                     onLongPressEnd: initInputDevice == InputDevice.touch ? _handleLongPressEnd : null,
-                    onVerticalDragStart: initInputDevice == InputDevice.touch ? _handleVerticalDragStart : null,
-                    onVerticalDragUpdate: initInputDevice == InputDevice.touch ? _handleVerticalDragUpdate : null,
-                    onVerticalDragEnd: initInputDevice == InputDevice.touch ? _handleVerticalDragEnd : null,
                   ),
                 ),
+                if (initInputDevice == InputDevice.touch)
+                  Positioned.fill(
+                    child: GestureDetector(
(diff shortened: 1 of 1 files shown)


lib/screens/video_player/video_player_controls.dart
```diff
@@ -135,11 +135,36 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
...
                 ),
+                if (initInputDevice == InputDevice.touch) ...[
+                  Align(
+                    alignment: Alignment.centerLeft,
+                    child: FractionallySizedBox(
+                      widthFactor: 0.15,
+                      heightFactor: 0.8,
+                      child: GestureDetector(
+                        behavior: HitTestBehavior.translucent,
+                        onVerticalDragStart: (details) => _handleVerticalDragStart(details, isLeft: true),
```
PartyDonut: This seems very hacky to have two GestureDetectors. 

We should be able to infer the location of the drag using the `DragStartDetails`

PartyDonut (changes requested):

Issue #429, opened 2025-08-02
🐛 some movie not forwarding


Gujjugaming2k:
### Description

some movie not forwarding 

lets say one movie is abc.mkv which is stored in .strm file
in mobile its working fine but in web its not forwarding its keep start from beginning only
resume as well not working

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Platform

Web

### App Version

latest

Issue #215, opened 2025-02-02
STRM working now, but after skip it's keep loading not playing


Gujjugaming2k:
### Description

Hi 

I can see as per last update strm is playing now but if we skip or resume from last it's keep loading and not playing check once

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
![DESCRIPTION](LINK.png)
```

### Logs

```bash

```

### Platform

Android

Gujjugaming2k:
@PartyDonut have you check this ?
have you tested skip and resume method if you need test I can provide server details to check

Mehulittv:
@PartyDonut  I have fix this Please checkout and if possible add in your code,  it will fetch strm link using download API 

Items/$itemId/Download?api_key=$apiKey

store that URL on getDirectDownloadUrl and use on fallbackUrl 

https://github.com/Mehulittv/Fladder/blob/develop/lib/models/playback/playback_model.dart

PartyDonut:
@Gujjugaming2k @Mehulittv Once again please refrain from tagging someone just to bump up the issue. Issues will get worked on when someone is willing/able.

To the issue at hand.

The fix itself is not released yet you need to wait for that before you can test #57 . 

> [@PartyDonut](https://github.com/PartyDonut) have you check this ? have you tested skip and resume method if you need test I can provide server details to check

Yes I tested it with the following .strm files both files play correctly including skip/resume
[StreamTestFiles.zip](https://github.com/user-attachments/files/18690804/StreamTestFiles.zip)

> [@PartyDonut](https://github.com/PartyDonut) I have fix this Please checkout and if possible add in your code, it will fetch strm link using download API
> 
> Items/$itemId/Download?api_key=$apiKey
> 
> store that URL on getDirectDownloadUrl and use on fallbackUrl
> 
> https://github.com/Mehulittv/Fladder/blob/develop/lib/models/playback/playback_model.dart

While this does work, its only a partial fix. 
Jellyfin only allows download urls to work if that user has download permissions. So this wil break if the server has this disabled for your users.

Check #210 for the original fix.

Also there is a lot of duplicate functionality for instance adding dio (not sure why?). That's fine but do note I can't test something if the behaviour changes from using your own fork of Fladder.

TL;DR: Closing this issue. Please keep any future comments or discussions in #57, as mentioned there, to avoid opening new tickets for the same issue.

Issue #159, opened 2024-11-20
after playing direct mode skip taking lots of time


Gujjugaming2k:
### Description

Hi Team,

once video started playing and if we skip timestamp its taking lots of time to start from there, I can share my temp id and server if you want to test it

we have stored all link on .strm file like https://abc.com/abc/mkv   this link will be stored on strm file as abc.strm

 

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....Hi Team,

once video started playing and if we skip timestamp its taking lots of time to start from there, I can share my temp id and server if you want to test it

we have stored all link on .strm file like https://abc.com/abc/mkv   this link will be stored on strm file as abc.strm


 '
4. See error
```

### Screenshots

```bash
![DESCRIPTION](LINK.png)
```

### Logs

```bash

```

### Platform

Android, Web

Gujjugaming2k:
let us know I can provide test login to check in details I can share logs as well if required ?

Issue #57, opened 2024-10-20
Direct Play STRM not working


Gujjugaming2k:
### Description

Have you tested STRM ?

which will content direct link, its not working 

### Screenshots

![DESCRIPTION](LINK.png)

### Platform

Web, Android, Windows

### App Version

latest

PartyDonut:
Hi, I could use some more information on this issue. 
VideoCodec,VideoContainer etc. anything related to playback and what is not working.

Remuxing does seem to work on my end when I orce it to remux from mkv to mp4. 
However where jellyfin reports remuxing, Fladder still show as transcoding.

Gujjugaming2k:
Title: Hoisting the Flag
Container: mkv
Path: /opt/jellyfin/STRM/VFlix_STRM/Series/AMZN/Suzhal The Vortex/S01/Suzhal The Vortex S01 E01 WebRip 720p Hindi  Tamil AAC 5.mkv.strm
Size: 598.72 MB

Title: 720p H264 SDR
Codec: h264
Profile: High
Level: 41.0
Resolution: 1280x720
Aspect Ration: 16:9
Interlaced: false
FrameRate: 25.0
Bitrate: 1486819 kbps
Bit depth: 8
Video range: VideoRange.sdr
Video range type: VideoRangeType.sdr
Ref frames: 1

Title: Hindi - AAC - 5.1 - Default
Language: hin
Codec: aac
Layout: 5.1
Bitrate: 191578 kbps
Sample Rate: 48000 Hz
Default: true
Forced: false
External: false
Title: Tamil - AAC - 5.1
Language: tam
Codec: aac
Layout: 5.1
Bitrate: 191578 kbps
Sample Rate: 48000 Hz
Default: false
Forced: false
External: false

Title: English [SDH] - Default - SUBRIP
Language: eng
Codec: subrip
Profile: null
Default: true
Forced: false
External: false
Title: English - SUBRIP
Language: eng
Codec: subrip
Profile: null
Default: false
Forced: false
External: false
Title: French - SUBRIP
Language: fre
Codec: subrip
Profile: null
Default: false
Forced: false
External: false
Title: Hindi - SUBRIP
Language: hin
Codec: subrip
Profile: null
Default: false
Forced: false
External: false
Title: Korean - SUBRIP
Language: kor
Codec: subrip
Profile: null
Default: false
Forced: false
External: false
Title: Dutch - SUBRIP
Language: dut
Codec: subrip
Profile: null
Default: false
[...]

PR #979, opened 2026-04-26
fix: ImagesData.firstOrNull throws RangeError on empty backDrop list

Beary-Handsome:
## What

Replace `backDrop?[0]` with `backDrop?.firstOrNull` (from `package:collection`, already imported in this file).

```diff
   ImageData? get firstOrNull {
-    return primary ?? backDrop?[0];
+    return primary ?? backDrop?.firstOrNull;
   }
```

## Why

When `primary` is null AND `backDrop` is an empty list (rather than null) — which happens for any Jellyfin item whose `backdropImageTags` is `[]` — `backDrop?[0]` indexes an empty list and throws:

```
RangeError (length): Invalid value: Valid value range is empty: 0
```

The getter is named `firstOrNull` but the implementation didn't actually handle the empty-list case. `package:collection`'s `firstOrNull` extension has the correct semantics.

## Symptom this fixes

In `MediaControlsWrapper.play`, `playBackItem.images?.firstOrNull` is evaluated synchronously. The unhandled RangeError escapes through the `await state.play()` chain in `VideoPlayerNotifier.loadPlaybackItem` and aborts video pipeline init. Audio that was already buffered keeps playing, but the video stream never starts decoding — externally this looks like "audio works but video keeps loading forever."

Reproducible on any non-Windows/non-Web build (i.e. Linux desktop, Android, iOS, macOS) playing an item where Jellyfin returns no backdrop images. The most reliable triggers are `.strm`-backed Episode items (e.g. content fed through a Usenet/Easynews bridge) — these typically have minimal Jellyfin metadata and no backdrop array — but any Episode/Movie with `backdropImageTags: []` and no primary image is affected.

Likely related: #57, #159, #215, #429 — all reported the symptom (playback hang on streamed/`.strm` content) but never landed on the root cause.

## How to reproduce

1. Set up a Jellyfin library where some items have `backdropImageTags: []` and a missing/null primary image. (Or run any Easynews-style `.strm` bridge.)
2. Build a non-Windows Fladder.
3. Play one of those items. Audio plays, video stays in "loading" indefinitely. Run from terminal and you'll see `Unhandled Exception: RangeError (length)` in stderr.

After this patch: the getter correctly returns null on empty, the synchronous evaluation no longer throws, and playback proceeds normally.

## Test plan

- [x] Reproduce the bug on Linux desktop with Easynews-backed episodes (audio plays, video stuck loading)
- [x] Apply patch, rebuild, retest — video plays through cleanly
- [x] Confirm `package:collection` import is already present in the file (it is, used by `randomBackDrop` on the next line)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -168,35 +168,25 @@ jobs:
       - name: Get dependencies
         run: flutter pub get
 
-      - name: Build Android APK (PRs)
-        if: needs.fetch-info.outputs.build_type == 'development'
-        run: |
-          flutter build apk --debug --split-per-abi --build-number=${{github.run_number}} --flavor production
-
-      - name: Build Android APK and AAB
-        if: needs.fetch-info.outputs.build_type != 'development'
+      - name: Build Android APK
         run: |
           flutter build apk --release --split-per-abi --build-number=${{github.run_number}} --flavor production
-          flutter build appbundle --release --build-number=${{github.run_number}} --flavor production
 
-      - name: Rename APK for PR
-        if: needs.fetch-info.outputs.build_type == 'development'
-        run: |
-          mkdir -p build/app/outputs/android_artifacts
-          # Move split debug APKs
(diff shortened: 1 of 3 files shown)


PartyDonut:
Already fixed in #969.

PR #986, opened 2026-05-09
fix: improve play/pause behavior on desktop

### LINKED ISSUE #580: Click to pause/play sometimes not working
### Description

On Windows when moving the mouse into the player window and clicking very fast sometimes does not pause/play. Waiting a second before clicking (letting the ui fade in finish) always works.

### Reproduction steps

```bash
Move mouse into player window and click very fast.
```

### Screenshots

_No response_

### Logs

```bash

```

### Platform

Windows

### App Version

v0.8.0

loregard735:
## Pull Request Description

 Changed the GestureDetector on video_player_controls.dart:

-  onTap now handles only touch logic
- onHorizontalDragDown handles pointer play/pause logic (previously on onTap), it triggers on clicks with mouse moving also, improving reliability.

## Issue Being Fixed
Before, play/pause on click would not register if mouse was still moving when the click event triggered. But mouse movement is to be expected when quickly switching screens/windows, etc.

solves  #580

## Screenshots / Recordings

old: play/pause fails when mouse moving

https://github.com/user-attachments/assets/353612ed-a11e-43b3-9162-33961e85659b

new: play/pause works with mouse moving

https://github.com/user-attachments/assets/76688926-650d-4968-8777-20e048a097b4

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [X] Windows
- [X] macOS
- [ ] Web

## Checklist

### DIFF
--- a/lib/screens/video_player/video_player_controls.dart
+++ b/lib/screens/video_player/video_player_controls.dart
@@ -128,7 +128,7 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
               children: [
                 Positioned.fill(
                   child: GestureDetector(
-                    onTap: initInputDevice == InputDevice.pointer ? () => player.playOrPause() : () => toggleOverlay(),
+                    onTap: initInputDevice == InputDevice.pointer ? null : () => toggleOverlay(),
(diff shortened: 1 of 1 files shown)


PartyDonut (approved):
Looks good to me and does work better then just onTap. Thanks 😄

PR #991, opened 2026-05-13
Display episode air date in video playback and controls

ijaron:
## Pull Request Description

This change adds episode airdate metadata to the player UI so it is visible during playback instead of only being inferred from episode labels.

#### What was changed
* Added airdate display to the playback details row in the video player controls.
* Added airdate display to the floating mini player subtitle line.
* Airdate is only shown for episode items and only when a valid airdate exists.
* Airdate formatting uses localized date formatting to match the user locale.

#### Result

When playing TV episodes, users now see the episode airdate directly in the active playback UI surfaces where episode context is shown.

## Issue Being Fixed

Episode playback UI was missing explicit airdate information in key player surfaces (full controls details row and floating player bar), making it harder to identify episode chronology at a glance. This update restores and standardizes airdate visibility in those locations.

## Screenshots / Recordings

image

image

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [x] Windows
- [ ] macOS
- [x] Web

## Checklist

### DIFF
--- a/lib/screens/video_player/components/video_playback_information.dart
+++ b/lib/screens/video_player/components/video_playback_information.dart
@@ -1,10 +1,12 @@
 import 'dart:ui';
 
 import 'package:flutter/material.dart';
+import 'package:intl/intl.dart';
 
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:iconsax_plus/iconsax_plus.dart';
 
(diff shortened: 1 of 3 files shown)


PartyDonut:
While I appreciate the changes, I don't really see a need for having the airdate visible in those places. 

Clutters the UI with information most people would have no use for.

PR #995, opened 2026-05-19
feat(pip): add Picture-in-Picture support for Android and iOS

aviadlevy:
Adds Picture-in-Picture (PiP) support for the Flutter-rendered video players (`libMPV` / `libMDK`) on Android, with the iOS plumbing in place but untested. The standalone native ExoPlayer `VideoPlayerActivity` is intentionally out of scope.

### Behavior

- A **PiP icon** is now part of the player controls (bottom row, landscape only — see "Decisions" below). Tapping it enters PiP at any time.
- A new **"Auto Picture-in-Picture"** toggle under *Settings → Player* controls whether the OS should also auto-enter PiP when the app is backgrounded from the player. Defaults to **on**.
- While in PiP, the controls overlay is hidden so only the video surface (plus the subtitle widget, if any) is captured for the PiP window.
- On player dispose, auto-enter is turned **off** so the user is never auto-PiP'd from unrelated screens.

### Architecture

- **`lib/wrappers/pip_manager.dart`** — new `PipManager` wrapper hides `package:pip` behind a narrow `PipClient` interface so it can be unit-tested without the native plugin. Exposes `enable` / `enter` / `disable` / `dispose` and a `Stream<bool>` for the current PiP state. 9 unit tests.
- **`lib/providers/pip_provider.dart`** — `pipManagerProvider` + `pipStateProvider`.
- **`VideoPlayer`** integration listens to the auto-enter preference and re-applies it live if the user toggles the setting mid-playback.

### Platform plumbing

- `pubspec.yaml`: add `pip: ^0.0.3` (only file in `lib/` that imports `package:pip` is the wrapper).
- `android/app/src/main/AndroidManifest.xml`: add `android:supportsPictureInPicture="true"` to `MainActivity`. `configChanges` already covers `screenSize|smallestScreenSize|screenLayout`.
- `ios/Runner/Info.plist`: add `audio` to `UIBackgroundModes` (required for iOS PiP per `package:pip` docs).

### Decisions taken during implementation

1. **Setting semantics split.** Originally the single toggle gated both manual button visibility *and* auto-enter. Per UX feedback this was split: the toggle now controls *only* the OS auto-enter-on-background behavior; the manual PiP button is always available on Android/iOS.
2. **Button placement.** Tried three positions: bottom-row always, top-bar after minimize button, top-bar before close-X. Landed on **bottom row, after the `…` more button, landscape-only** — phone portrait was overflowing the row (RIGHT OVERFLOWED BY 46 PIXELS). Portrait users access PiP via the Home button + the auto-enter setting.
3. **Icon.** Material `Icons.picture_in_picture_alt_outlined` (rather than an IconsaxPlusLinear equivalent) — matches the de-facto Flutter standard used by other video apps (frosty, mydia) and YouTube/Netflix; the universal "rectangle-with-corner-rectangle" glyph is more discoverable than a generic maximize icon.
4. **Aspect ratio hardcoded 16:9.** `PlayerState` does not currently expose video width/height. Plumbing real ratios through every backend was deferred to keep the PR focused.
5. **Native player out of scope.** The separate `VideoPlayerActivity` (ExoPlayer/Compose) already has `supportsPictureInPicture` in its manifest but no logic; wiring that needs a separate Kotlin-side change and would have doubled the PR's surface area.
6. **One-pref-toggle design** rather than the three-state `Never / Always / Only with headphones` from the original discussion. Smaller surface, easier to maintain; can be expanded later if there's demand.

### Caveats & known limitations

- **iOS untested.** The Dart code is cross-platform and the iOS-side `Info.plist` change is in, but I do not have an iOS dev environment. The flow *should* work per `package:pip` docs but should be validated by someone with an iPhone before being treated as supported.
- **Aspect ratio is fixed at 16:9** as noted above. Non-16:9 videos will have black bars inside the PiP frame on Android.
- **Subtitle rendering inside the PiP frame** depends on the backend. Flutter-rendered subtitles are captured; subtitles burned into the player's native surface inherit the surface and also appear. No special handling was added beyond keeping the subtitle widget visible during PiP.

### Implementation notes

This PR was implemented with the help of an AI coding assistant (Claude Code). All decisions, UX trade-offs, and code review were human-driven; the AI handled mechanical edits and the wrapper/test scaffolding. Generated code follows the project's existing conventions (`dart format --line-length 120`, sparse comments, freezed for settings).

## Issue Being Fixed

Picture-in-Picture has been requested multiple times across several discussions. This PR addresses the **mobile (Android)** portion of those asks.

- [Discussion #494 — *Allow PiP for Mobile*](https://github.com/DonutWare/Fladder/discussions/494): the most direct ask; explicitly mobile-focused. **Fully addressed.**
- [Discussion #738 — *Double Tap to skip and PIP*](https://github.com/DonutWare/Fladder/discussions/738): combined request. The double-tap-to-seek part is already implemented (`enableDoubleTapSeek` setting). The PiP part — assumed mobile based on context — **is addressed**.

## Screenshots / Recordings

<table>
  <tr>
    <td>Settings — Auto PiP toggle</td>
    <td>Portrait player — no overflow</td>
    <td>In PiP window</td>
  </tr>
  <tr>
    <td colspan="3">Landscape player — PiP icon in bottom row</td>
  </tr>
</table>

## Tested On

- [x] Android — Pixel 9 Pro XL emulator (API 34) and Pixel 10 Pro XL hardware (Android 16, API 36). All four scenarios verified: manual button, auto-enter on Home, toggle off → no auto-PiP, close PiP via × → playback ends and resume position is reported to Jellyfin.
- [ ] Android TV
- [ ] iOS — code in place, not validated on device.
- [ ] Linux — N/A (no PiP on desktop)
- [ ] Windows — N/A
- [ ] macOS — N/A
- [ ] Web — N/A

## Checklist

- [x] If a new package was added, did you ensure it works for all supported platforms? Is the package well maintained — `pip ^0.0.3` is a 0.0.x release with limited maintenance history; flagging as a known trade-off. Cross-platform (Android/iOS) by design. Open to swapping for a more mature alternative if maintainers prefer.

### DIFF
--- a/lib/screens/video_player/video_player_controls.dart
+++ b/lib/screens/video_player/video_player_controls.dart
@@ -382,7 +382,7 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
                                 );
                               }
                             },
-                            icon: const Icon(Icons.picture_in_picture_alt_outlined),
+                            icon: const Icon(IconsaxPlusLinear.screenmirroring),
                           ),
                         if (AdaptiveLayout.layoutOf(context) == ViewSize.tablet) ...[
                           IconButton(
--- a/lib/screens/video_player/video_player_controls.dart
+++ b/lib/screens/video_player/video_player_controls.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
@@ -16,6 +17,7 @@ import 'package:fladder/models/items/media_streams_model.dart';
 import 'package:fladder/models/media_playback_model.dart';
 import 'package:fladder/models/playback/playback_model.dart';
 import 'package:fladder/models/settings/video_player_settings.dart';
+import 'package:fladder/providers/pip_provider.dart';
 import 'package:fladder/providers/settings/client_settings_provider.dart';
 import 'package:fladder/providers/settings/video_player_settings_provider.dart';
 import 'package:fladder/providers/user_provider.dart';
@@ -96,8 +98,18 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
 
   @override
   Widget build(BuildContext context) {
-    final mediaSegments = ref.watch(playBackModel.select((value) => value?.mediaSegments));
+    final isInPip = ref.watch(pipStateProvider).asData?.value ?? false;
     final player = ref.watch(videoPlayerProvider);
+    if (isInPip) {
+      // Keep only the subtitle widget so it's captured in the PiP frame.
+      final pipSubtitleWidget = player.subtitleWidget(false, controlsKey: _bottomControlsKey);
+      return Stack(
+        children: [
+          if (pipSubtitleWidget != null) Positioned.fill(child: pipSubtitleWidget),
+        ],
+      );
+    }
+    final mediaSegments = ref.watch(playBackModel.select((value) => value?.mediaSegments));
     final subtitleWidget = player.subtitleWidget(showOverlay, controlsKey: _bottomControlsKey);
     final isDesktop = AdaptiveLayout.of(context).isDesktop || kIsWeb;
     final speedBoostEnabled = ref.watch(videoPlayerSettingsProvider.select((value) => value.enableSpeedBoost));
@@ -357,6 +369,21 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
                         IconButton(
                             onPressed: () => showVideoPlayerOptions(context, () => minimizePlayer(context)),
                             icon: const Icon(IconsaxPlusLinear.more)),
+                        if (!kIsWeb &&
+                            (Platform.isAndroid || Platform.isIOS) &&
+                            MediaQuery.orientationOf(context) == Orientation.landscape)
+                          IconButton(
+                            tooltip: context.localized.pictureInPictureTitle,
+                            onPressed: () async {
+                              final ok = await ref.read(pipManagerProvider).enter();
+                              if (!ok && context.mounted) {
+                                ScaffoldMessenger.of(context).showSnackBar(
+                                  SnackBar(content: Text(context.localized.pictureInPictureNotSupported)),
+                                );
+                              }
+                            },
+                            icon: const Icon(Icons.picture_in_picture_alt_outlined),
+                          ),
                         if (AdaptiveLayout.layoutOf(context) == ViewSize.tablet) ...[
                           IconButton(
                             onPressed: () => showSubSelection(context),
--- a/lib/screens/video_player/video_player_controls.dart
+++ b/lib/screens/video_player/video_player_controls.dart
@@ -369,8 +369,7 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
                         IconButton(
                             onPressed: () => showVideoPlayerOptions(context, () => minimizePlayer(context)),
                             icon: const Icon(IconsaxPlusLinear.more)),
-                        if (pipPlatformSupported &&
-                            MediaQuery.orientationOf(context) == Orientation.landscape)
+                        if (pipPlatformSupported && MediaQuery.orientationOf(context) == Orientation.landscape)
                           IconButton(
                             tooltip: context.localized.pictureInPictureTitle,
                             onPressed: () async {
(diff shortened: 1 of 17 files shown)


aviadlevy:
Found a small thing I missed - when minimized and then home is pressed it's not getting into pip, just exit.   
Expected behavior - should be pip mode

I'll be able to tackle it on Sunday

GustavoGarufi:
This looks like an awesome feature. I'm looking forward to seeing this live!

aviadlevy:
> This looks like an awesome feature. I'm looking forward to seeing this live!

I'm using this PR version for a while now without issues, and it's really the thing I missed on the official version (watching mainly on the road with some context switch while watching)

lib/wrappers/pip_manager.dart
```diff
@@ -0,0 +1,148 @@
+import 'dart:async';
+import 'dart:io';
+
+import 'package:flutter/foundation.dart';
+import 'package:pip/pip.dart';
+
+bool _platformSupportsPip() {
+  if (kIsWeb) return false;
```
PartyDonut, line 8: This logic/check seems to be duplicated in a few places. Would it not be better to use this function everywhere we need to check for Pip support?

PartyDonut (changes requested):
Thanks for having a go at implementing this, looks and works great. One small change request, after that we can merge it.

lib/screens/video_player/video_player_controls.dart
```diff
@@ -350,6 +362,21 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
...
+                            tooltip: context.localized.pictureInPictureTitle,
+                            onPressed: () async {
+                              final ok = await ref.read(pipManagerProvider).enter();
+                              if (!ok && context.mounted) {
+                                ScaffoldMessenger.of(context).showSnackBar(
+                                  SnackBar(content: Text(context.localized.pictureInPictureNotSupported)),
+                                );
+                              }
+                            },
+                            icon: const Icon(Icons.picture_in_picture_alt_outlined),
```
PartyDonut: Lets use "IconsaxPlusLinear.screenmirroring" for this to keep the icons in the same style

aviadlevy: Addressed in https://github.com/DonutWare/Fladder/pull/995/changes/0fb4b871370883c19b09c373f5c5a44592d50d4b

lib/wrappers/pip_manager.dart
aviadlevy: Addressed in https://github.com/DonutWare/Fladder/pull/995/changes/ca6aa9bb5962d01151009613a763cd7f5cdf6060  

Hope this is what you meant

PartyDonut (approved):
Looks good to me thanks for the quick fixes

aviadlevy:
> Looks good to me thanks for the quick fixes

Thank you for this amazing app! It replaced all my previous android & tv & desktop apps for Jellyfin!  

When should we expect a new version?

PartyDonut:
> > Looks good to me thanks for the quick fixes
> 
> Thank you for this amazing app! It replaced all my previous android & tv & desktop apps for Jellyfin!
> 
> When should we expect a new version?

Thanks good to hear you like the application 😄 

Release will be in the next few days probably, depends on how quickly I can fix the small remaining issues with the music player and offline playback

schembriaiden:
@aviadlevy thanks for this addition. One thing I noticed now when testing the latest nightly is that when you are playing episodes and the next up screen comes up and you are in PiP it bugs out and just show a black screen. Could we maybe get a small button like the "Skip" button in the Youtube PiP mode?

This is the issue:

Screenshot_20260821_124202_One UI Home

The proposed solution:

image

It seems currently like it just shrinks the whole screen instead, maybe we could try keeping the current episode playing and it just shows the next button when the episode is about to end in PiP?

PartyDonut:
@aviadlevy 
Probably better to open a new issue so it can be tracked if possible. Might get lost if the issue is inside a merged pull-request.

aviadlevy:
Agree, probably best if the one reporting it will open the new issue 😄 
FYI @schembriaiden

schembriaiden:
> Agree, probably best if the one reporting it will open the new issue 😄 FYI @schembriaiden

Yep opening it now. I posted here first because it would be easier to get to your attention due to the notifications 😆 

https://github.com/DonutWare/Fladder/issues/1096

PartyDonut:
> Agree, probably best if the one reporting it will open the new issue 😄 

Haha oops copied the wrong @ when replying, my bad. Thanks for the correction 😃

PR #997, opened 2026-05-23
feat(Android): Implement refresh rate matching for native player

TwoToneEddy:
## Pull Request Description

Ports refresh rate matching from the native Android TV client to Fladder's native player backend on Android. When a video begins playing, the display refresh rate is switched to match the video's frame rate using `Surface.setFrameRate()` / `Window` preferred display mode. Playback pauses during the mode switch and resumes automatically once the display has settled. A toggle to enable/disable the feature is added to player settings.

Also fixes a visual artifact on player exit where the video texture was released mid-frame during the pop animation, causing the top-left quadrant of the video to flash briefly in the bottom-right of the UI. `closePlayer()` now awaits `endOfFrame` before navigating so the widget tree can rebuild and cleanly detach the texture first.

## Tested On
- [x] Android
- [x] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -1,3 +1,7 @@
 org.gradle.jvmargs=-Xmx4G -XX:MaxMetaspaceSize=2G -XX:+HeapDumpOnOutOfMemoryError
 android.useAndroidX=true
 android.enableJetifier=true
+# This builtInKotlin flag was added automatically by Flutter migrator
+android.builtInKotlin=false
+# This newDsl flag was added automatically by Flutter migrator
+android.newDsl=false
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -1,7 +1,3 @@
 org.gradle.jvmargs=-Xmx4G -XX:MaxMetaspaceSize=2G -XX:+HeapDumpOnOutOfMemoryError
 android.useAndroidX=true
 android.enableJetifier=true
-# This builtInKotlin flag was added automatically by Flutter migrator
-android.builtInKotlin=false
-# This newDsl flag was added automatically by Flutter migrator
-android.newDsl=false
--- a/android/app/src/main/kotlin/nl/jknaapen/fladder/VideoPlayerActivity.kt
+++ b/android/app/src/main/kotlin/nl/jknaapen/fladder/VideoPlayerActivity.kt
@@ -1,6 +1,7 @@
 package nl.jknaapen.fladder
 
 import android.graphics.PixelFormat
+import android.hardware.display.DisplayManager
 import android.os.Build
 import android.os.Bundle
 import android.view.WindowManager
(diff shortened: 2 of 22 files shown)


android/gradle.properties
```diff
@@ -1,3 +1,7 @@
 org.gradle.jvmargs=-Xmx4G -XX:MaxMetaspaceSize=2G -XX:+HeapDumpOnOutOfMemoryError
 android.useAndroidX=true
 android.enableJetifier=true
+# This builtInKotlin flag was added automatically by Flutter migrator
+android.builtInKotlin=false
```
PartyDonut: I think these might be added because the flutter version you use is different from the version in .fvmrc? Probably best to remove this if we are unsure what they do exactly.

PartyDonut (changes requested):
Thanks for implementing this, looks good and seems to work.

I will admit not sure if I should see a visual change on TV but that might just be a me problem.

If you could re-base as well that would be great 👍

TwoToneEddy:
Thanks for taking a look. I have since added a soundbar to my system which passes through the HDMI. Since using this rather than direct to the TV I am noticing some issues with what I believe to be handshaking. I will investigate this over the next week an hopefully fix, along with your suggestion. Thanks

android/gradle.properties
TwoToneEddy: Good catch — you're right about where they came from. Before settling on the version pinned in .fvmrc I briefly had a newer Flutter checkout pointed at the project, and its migrator inserted those two flags automatically. They're not referenced by the Gradle plugin in 3.35.7 at all (built-in Kotlin / new DSL are from newer Flutter releases), so they were inert here.

I've removed them and rebuilt the production APK with 3.35.7 to confirm nothing depends on them — build passes. Pushed as a follow-up commit.

TwoToneEddy:
Seems to work reliably now with my soundbar so OK to merge once we sort those conflicts

TwoToneEddy:
OK i've done a rebase, everything seems to build and work still. Hopefully this is all OK

PartyDonut (approved):
Looks good to me thanks for implementing this. You do seem to have claude co-authored in the 2 latest commits. While I don't mind using AI, the co-authoring label is something I would like to prevent. If you could clean that up that would be great.

PartyDonut (approved):

PAXANDDOS:
Has there been any progress on getting this delivered? Would be great to have refresh-rate matching in Fladder TV app

TwoToneEddy:
Ah apologies I forgot  I had to resolve these conflicts. I'll do it at lunch today. Sorry for the delay, I agree it would be great to have it in the next release

TwoToneEddy:
FYI that last commit was me

Issue #966, opened 2026-04-18
Fladder flatpak runs on org.gnome.Platform branch 48 to be deprecated march 24, 2026


EldoDvae:
### Description

Upon hitting `flatpak update -i`, I was hit with the following message:

❯ flatpak update -y
Looking for updates…

Info: runtime org.gnome.Platform branch 48 is end-of-life, with reason:
   The GNOME 48 runtime is no longer supported as of March 24, 2026. Please ask your application developer to migrate to a supported platform.
Info: applications using this runtime:
   nl.jknaapen.fladder

Using fladder 0.10.3 flatpak.

### Reproduction steps

```bash
1. Update flatpak through command flatpak update
2. Message will be recieved:
Info: runtime org.gnome.Platform branch 48 is end-of-life, with reason:
   The GNOME 48 runtime is no longer supported as of March 24, 2026. Please ask your application developer to migrate to a supported platform.
Info: applications using this runtime:
   nl.jknaapen.fladder
```

### Logs

```bash

```

### Platform

Linux

### App Version

0.10.3 (630)

### Jellyfin server

10.11.7

schembriaiden:
We are in the process of releasing this app on flatpak itself, when the package is released there I will update the package here too.

EldoDvae:
Oh, nice! It would be nice to have it on flathub, soon.

PR #1012, opened 2026-06-02
chore: flathub branding and updated documentation

schembriaiden:
## Pull Request Description

This PR contains installation instructions for Flathub and branding changes to the metainfo.xml for Flathub so that the app has all the required qualities:

image

Maybe you could take a look at these and give me some suggestions on what we should do?

https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#reasonable-footprint

https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#in-line-with-contemporary-styles

https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#not-technical

https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#dont-start-with-an-article

https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#default-settings

https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#include-window-shadow-and-rounded-corners

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -7,6 +7,7 @@ Platform-specific installation instructions can be found in this document.
 - [Windows](#windows)
 - [macOS](#macos)
 - [Linux](#linux)
+	- [Flatpak](#flatpak)
 	- [Ubuntu/Debian](#ubuntudebian)
 	- [Arch](#arch)
 	- [Fedora](#fedora)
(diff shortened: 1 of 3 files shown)


schembriaiden:
I was also thinking maybe we should remove the flatpak builds we have here and make everything go through flathub. I think its a bit of a pain to maintain both at the same time. I don't know if you think we should do that?

PartyDonut (approved):
Looks good thanks for fixing this. Also thanks for all the work you have done on getting it published on Flathub 😄

PartyDonut:
> I was also thinking maybe we should remove the flatpak builds we have here and make everything go through flathub. I think its a bit of a pain to maintain both at the same time. I don't know if you think we should do that?

We probably could as most users would download it from flathub itself. Lets keep it for now though, best to remove it when it really does become a problem.

schembriaiden:
> We probably could as most users would download it from flathub itself. Lets keep it for now though, best to remove it when it really does become a problem.

I was only saying because Gnome 48 is EOL anyways (https://github.com/DonutWare/Fladder/issues/966), and instead of updating it I was thinking we just remove it, but if you want to keep it for now that's fine with me.

PR #1022, opened 2026-06-09
fix: play synced items without a reachable server (#970)

### LINKED ISSUE #713: 🐛 Offline playback not working on android
### Description

I cannot get offline playback to work on android, I am just left with a spinner.
I'm happy to provide screenshots if it'll be helpful.
Sorry if this is a skill issue!

### Reproduction steps

```bash
1. Open app without internet connection
2. Click on the syncs tab (cloud) on the bottom right
3. Press the 3 dots on the right of the synced series
4. Click on a season where there is a synced episode
5. Find a downloaded episode (I see 'Total size' in green and a delete icon on the right
6. Click on the thumbnail of the episode
7. Takes me to a grey screen with an spinner
```

### Screenshots

_No response_

### Logs

```bash

```

### Platform

Android

### App Version

0.9.0

### Jellyfin server

10.11.5 on podman

### LINKED ISSUE #970: Offline Synced data loading time very slow
### Description

When browsing the synced shows and movies offline (airplane mode on) on version 0.10.3 vs 0.10.2 the loading times are significantly greater. Same effect on both Android and IOS. Also watch history is bugged when watching offline and clicking to next episode of a show when the episode is complete, when done watching, no watch history has progressed. Have to manually input each episode as watched, paired with slow loading time is very inconvenient 

### Reproduction steps

```bash
1. Sync a show or a movie
2. turn on airplane mode and disconnect from the server
3. try to browse the synced shows
4. experience issue

The app seems to be trying to connect to the server before loading synced data despite being disconnected and showing offline on top
```

### Screenshots

_No response_

### Logs

```bash

```

### Platform

Android

### App Version

0.10.3

### Jellyfin server

10.11.2

aviadlevy:
Resolve sync status and connectivity before the server item-fetch in `createPlaybackModel`, via a new pure `useLocalSyncedCopy` predicate. When offline, skip the server and play the local synced copy immediately. Otherwise bound the fetch with a 5s timeout and fall back to the local copy when the server is unreachable (e.g. LAN-only server while on cellular). Online, server-reachable behavior is unchanged.

Fixes #970 
Fixes #713

## Pull Request Description

Playing **synced** items previously always contacted the server *before* checking whether the item was available locally. `createPlaybackModel` awaited `usersUserIdItemsItemIdGet` unconditionally, and that call masks failures internally (returning the local item only after the underlying request hangs to its OS socket timeout). The result: browsing and playing synced content while the server was unreachable was very slow, which is the core complaint in #970.

### Behavior

- Playing a synced item no longer needs the server. When the device is **offline**, playback starts directly from the local file with **no server round-trip**.
- When there **is** a network but the server is unreachable (e.g. a LAN-only server while commuting on cellular), the server item-fetch is bounded by a **5s timeout** and falls back to the local copy.
- Auto-advance to the next synced episode goes through the same path, so it no longer hangs offline either.
- **Online with a reachable server is unchanged** — the existing quality-selection dialog (Direct Stream / Transcode / Offline) still appears for synced items, and non-synced playback is untouched.

### Implementation

- `lib/models/playback/playback_model.dart`: `createPlaybackModel` now resolves the synced item and connectivity *before* the server fetch. A new pure predicate `useLocalSyncedCopy({required bool isSynced, required bool serverReachable})` (`isSynced && !serverReachable`) centralizes the source decision and is unit-tested in isolation. The fetch is wrapped with a 5s `.timeout(...)` (matching the existing reachability-check value) and falls back to the local copy when the server can't be reached.
- The previously-duplicated synced-item lookup was removed; the lookup is now keyed on `firstItemToPlay.id` so a series/season resolves its actual resume episode.
- `test/models/playback/playback_model_test.dart`: 4 unit tests for the predicate.

The "re-sync offline progress when back online" behavior already exists (`SyncNotifier` reacts to connectivity returning and flushes pending user-data), so no change was needed there.

### Implementation notes

Implemented with the help of an AI coding assistant (Claude Code); all decisions, trade-offs, and review were human-driven. Generated code follows the project's conventions.

## Issue Being Fixed

Resolves #970

Synced shows/movies load very slowly when the server isn't reachable because the app tries to contact the server before falling back to the local copy. This PR removes that dependency for synced playback. (The issue also mentions watch-history not advancing offline; offline progress is saved locally and pushed on reconnect via the existing sync path — worth confirming during testing.)

## Tested On

- [x] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

(but this change applicable for all who supporting offline sync)

## Checklist

- [x] If a new package was added, did you ensure it works for all supported platforms? Is the package well maintained — no new packages added.

### DIFF
--- a/lib/models/playback/playback_model.dart
+++ b/lib/models/playback/playback_model.dart
@@ -154,6 +154,9 @@ final playbackModelHelper = Provider<PlaybackModelHelper>((ref) {
   return PlaybackModelHelper(ref: ref);
 });
 
+@visibleForTesting
+bool useLocalSyncedCopy({required bool isSynced, required bool serverReachable}) => isSynced && !serverReachable;
+
 class PlaybackModelHelper {
   const PlaybackModelHelper({required this.ref});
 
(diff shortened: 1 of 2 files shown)


aviadlevy:
This is probably covered by https://github.com/DonutWare/Fladder/pull/1030
I'll close this. Thanks @PartyDonut for the more robust solution

PR #1033, opened 2026-06-17
fix(macOS): Double click titlebar to zoom

aaravjhamb:
The window uses a full-size content view with a hidden title bar, so the Flutter view covers the entire window and AppKit's default "double-click the title bar to zoom/minimize" never reaches the native title bar.

## Pull Request Description

Added a proper macos titlebar that supports macOS actions like double clicking to snap
## Issue Being Fixed

Resolves the issue of titlebar not snapping to fullscreen on macOS

## Screenshots / Recordings

image

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [x] macOS
- [ ] Web

## Checklist

- [X ] Check that any changes are related to the issue at hand.


PartyDonut (approved):
Thanks looks good. 

Was always wondering why the double click didn't work. But never had a moment to fix it.

PR #1044, opened 2026-07-03
fix: Subtitle/audio mismatch under certain conditions and with multiple default tracks

### LINKED ISSUE #1031: Incorrect subtitle track played when file has multipe default tracks
### Description

When a file has multiple subtitle tracks marked as default the UI seems to select the correct one according to my settings in jellyfin. However when I click play instead of the subtitle track the UI shows it's using it's actually using the first default track instead. 

I have tested this by removing the other default markers from the subtitle tracks and then it does select the correct track. 

### Reproduction steps

```bash
1. Have a movie/episode with multiple subtitle tracks marked as default.
2. Set your automatic subtitle selection settings in jellyfin so that the second default track gets selected by jellyfin.
example: I have "Preferred subtitle language" set to English and "Subtitle mode" set to smart. this way it selects the english subtitle track when the audio language is in a foreign language for me.
3. Play the movie/episode and it should set the correct track but play the first default track.
```

### Screenshots

In the movie/episode view the correct track is selected
Image

in the movie the correct track is selected but shows japanese subtitles
Image

after clicking the english track in the movie again
Image

my subtitle settings
Image

### Logs

```bash

```

### Platform

Linux

### App Version

0.10.3

### Jellyfin server

10.11.11

aviadlevy:
## Pull Request Description

At playback start, `setSubtitleTrack` / `setAudioTrack` (in `LibMPV`) ran before `media_kit` had finished parsing the file's tracks — `open()` returns before the mpv `track-list` observer fires. Both wrappers map the wanted Jellyfin stream to a player track **by list position**, so while the track list is still empty the lookup resolves to `null` and the selection is silently skipped. mpv then keeps the track it auto-selected on load — its first **default**-flagged track — which is wrong whenever a file has multiple default tracks (e.g. Jellyfin's smart subtitle selection picks the 2nd default).

This is exactly why the in-player UI showed the correct track while a different one actually played, and why re-selecting the same track fixed it (by then the track list had been parsed).

### Fix

`lib/wrappers/players/lib_mpv.dart`: before mapping the wanted stream to a player track, wait for that track to appear on `player.stream.tracks`. The wait:

- **only fires when the track isn't present yet** — already-loaded selections (manual re-selects, transcode) hit the fast path with zero added latency;
- **doesn't affect video load time** — the video is already playing (`play: true`), only the track application waits, and only until mpv finishes parsing (tens of ms);
- **is race-free** — Dart is single-threaded, so `firstWhere` subscribes synchronously before any track-list observer callback can run;
- **is capped at 5s** as a pure safety net, and **is skipped for external subtitles** (that path is unchanged).

Applied to both subtitle and audio selection. The `libMDK` wrapper has the same latent race but is intentionally left out of scope to keep the diff minimal.

### AI assistance

Root-cause investigation and the patch were done with Claude Code. The diagnosis — traced through the vendored DonutWare `media_kit` fork to confirm `open()` returns before track parsing and that `Tracks.subtitle` defaults to `[auto, no]` — the fix design, and on-device validation were human-reviewed and -owned.

## Issue Being Fixed

Resolves #1031 — when a file has multiple subtitle tracks marked as default, the UI selects the correct one but playback uses the first default track instead.

## Tested On

- [x] Android — installed the dev build on a physical device; confirmed the correct (non-first) default subtitle track now plays from the start without needing to re-select it.
- [ ] Android TV — uses the native ExoPlayer path (selection by index, not the positional remap), so it isn't affected by this bug.
- [ ] iOS
- [ ] Linux — same `LibMPV` code path as Android; not separately re-tested but covered by the same fix.
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

- [x] If a new package was added, did you ensure it works for all supported platforms? — no new packages.

### DIFF
--- a/lib/wrappers/players/lib_mpv.dart
+++ b/lib/wrappers/players/lib_mpv.dart
@@ -426,6 +426,15 @@ class LibMPV extends BasePlayer {
   @override
   Future<void> seek(Duration position) async => _player?.seek(position);
 
+  // mpv parses tracks asynchronously after open(); at playback start the list is still empty, so a
+  // positional lookup would miss and leave mpv on its own default pick. Wait (capped) for [index].
+  Future<void> _awaitTrack(int index, int Function(mpv.Tracks) count) async {
+    if (_player == null || index < 0 || count(_player!.state.tracks) > index + 2) return;
+    await _player!.stream.tracks
+        .firstWhere((tracks) => count(tracks) > index + 2)
+        .timeout(const Duration(seconds: 5), onTimeout: () => _player!.state.tracks);
+  }
+
   @override
   Future<int> setAudioTrack(AudioStreamModel? model, PlaybackModel playbackModel) async {
     final wantedAudioStream = model ?? playbackModel.defaultAudioStream;
@@ -433,9 +442,10 @@ class LibMPV extends BasePlayer {
     if (wantedAudioStream.index == AudioStreamModel.no().index) {
       await _player?.setAudioTrack(mpv.AudioTrack.no());
     } else {
+      final index = (playbackModel.audioStreams?.indexOf(wantedAudioStream) ?? -1) - 1;
+      await _awaitTrack(index, (tracks) => tracks.audio.length);
       final internalTracks = audioTracks.getRange(2, audioTracks.length).toList();
-      final audioTrack =
-          internalTracks.elementAtOrNull((playbackModel.audioStreams?.indexOf(wantedAudioStream) ?? -1) - 1);
+      final audioTrack = internalTracks.elementAtOrNull(index);
       if (audioTrack != null) {
         await _player?.setAudioTrack(audioTrack);
       }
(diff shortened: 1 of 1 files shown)


lib/wrappers/players/lib_mpv.dart
```diff
@@ -426,16 +426,26 @@ class LibMPV extends BasePlayer {
   @override
   Future<void> seek(Duration position) async => _player?.seek(position);
 
+  // mpv parses tracks asynchronously after open(); at playback start the list is still empty, so a
+  // positional lookup would miss and leave mpv on its own default pick. Wait (capped) for [index].
+  Future<void> _awaitTrack(int index, int Function(mpv.Tracks) count) async {
+    if (_player == null || index < 0 || count(_player!.state.tracks) > index + 2) return;
+    await _player!.stream.tracks
```
PartyDonut: Lets remove both bang operators.

While the first one is safe. When/if the timeout occurs and the player is null it wil return an exception instead of the state.tracks.

PartyDonut (changes requested):
Logic looks good. Left a small request

lib/wrappers/players/lib_mpv.dart
aviadlevy: Done in 10090bc7

PartyDonut (approved):
lgtm, thanks for the quick changes.

PR #1047, opened 2026-07-03
fix(Android): Keep screen awake after returning to video playback from background

### LINKED ISSUE #1046: Screen sleeps during video playback after returning from background
### Description

On Android, the screen turns off after the normal idle timeout **while a video is still playing**, instead of staying awake. It's intermittent — it does **not** happen if you just start a video and leave the app untouched. It reliably triggers only **after the app has been backgrounded and reopened** at least once during playback (e.g. pressing Home and returning, lock/unlock, or glancing at a notification). After that, if you set the phone down, the screen dims and sleeps at the system screen-off timeout even though playback continues.

### Reproduction steps

```bash
1. Start playing any video (default player).
2. Press Home to background the app, then reopen it (or lock/unlock, or dismiss a notification) — playback continues.
3. Set the phone down; do not touch the screen.
4. Wait for the system screen-off timeout (e.g. 60s).
5. The screen dims and turns off while the video is still playing.

Note: does NOT reproduce if you start a video and never leave the app —
only after at least one background→foreground round-trip.
```

### Screenshots

_No response_

### Logs

```bash

```

### Platform

Android

### App Version

_No response_

### Jellyfin server

10.11.11

aviadlevy:
## Pull Request Description

Fixes the screen sleeping mid-playback on Android (see #1046 for the full repro and on-device evidence).

Root cause: screen-on relied on the `FLAG_KEEP_SCREEN_ON` window flag toggled once in `play()`/`pause()`. Android drops that flag across a background→foreground cycle and nothing re-applied it — the resume handler meant to recover it (`if (playing) play()`) was dead code, since `playing` was never set `true`.

### Fix

- `media_control_wrapper.dart`: route the wakelock through one idempotent path — `_shouldKeepScreenOn(playing)` (on when playing and not an `AudioModel`) applied via `_applyWakelock(...)`, re-evaluated on every player state change instead of the scattered `WakelockPlus` calls.
- Added `reassertWakelock()` which **force**-re-applies the flag (our cached state reads "on" after the OS clears it, so a plain reconcile would skip the re-apply).
- `video_player.dart`: removed the dead `playing` flag; call `reassertWakelock()` on `AppLifecycleState.resumed`. Desktop/web stay excluded, as before.

The same handler also had a dead `pause()`-on-background branch (also gated on the never-true `playing`), which is removed too — behavior is unchanged, since playback already continued in the background. Wiring up actual background-pause would be a separate change.

### Scope

Native ExoPlayer path (Android TV) manages the flag separately in `ExoPlayer.kt` and is out of scope. No new packages.

### AI assistance

Root-cause investigation and the patch were done with Claude Code; diagnosis, fix design, and on-device validation were human-reviewed and -owned.

## Issue Being Fixed

Resolves #1046

## Tested On

- [x] Android — dev build on a physical Pixel 10 Pro XL (Android 17): reproduced the mid-playback sleep via a background→foreground cycle, then confirmed the keep-screen-on flag stays held and the screen stays awake through the idle timeout while the video keeps playing.
- [ ] Android TV — native ExoPlayer path, out of scope.
- [ ] iOS
- [ ] Linux — same `LibMPV`/`WakelockPlus` path as Android; not separately re-tested.
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

- [x] If a new package was added, did you ensure it works for all supported platforms? Is the package well maintained — no new packages added.

### DIFF
--- a/lib/wrappers/media_control_wrapper.dart
+++ b/lib/wrappers/media_control_wrapper.dart
@@ -74,8 +74,6 @@ class MediaControlsWrapper extends BaseAudioHandler implements VideoPlayerContro
   bool _isNewPlayback = false;
   bool _isAudioQueueMode = false;
   bool _audioQueueTransitioning = false;
-
-  /// Last requested wakelock state; used to skip redundant calls during steady playback.
   bool _wakelockEnabled = false;
 
   AudioPrefetchBuffer? _prefetchBuffer;
@@ -317,15 +315,13 @@ class MediaControlsWrapper extends BaseAudioHandler implements VideoPlayerContro
     return loadPreviousVideo();
   }
 
-  /// Video needs the screen awake; audio can keep playing with the screen off.
   bool _shouldKeepScreenOn(bool playing) {
     final item = ref.read(playBackModel.select((value) => value?.item));
     return playing && item is! AudioModel;
   }
 
-  /// Single source of truth for the wakelock. Idempotent during steady
-  /// playback; pass [force] to re-apply even when [_wakelockEnabled] already
-  /// matches (needed after Android silently clears the window flag on resume).
+  /// [force] re-applies even when the cached state already matches, since
+  /// Android silently clears the keep-screen-on flag while we still think it's set.
   Future<void> _applyWakelock(bool shouldEnable, {bool force = false}) async {
     if (!force && shouldEnable == _wakelockEnabled) return;
     _wakelockEnabled = shouldEnable;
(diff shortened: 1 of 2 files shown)


lib/wrappers/media_control_wrapper.dart
```diff
@@ -313,6 +317,32 @@ class MediaControlsWrapper extends BaseAudioHandler implements VideoPlayerContro
...
+      await WakelockPlus.enable();
+    } else {
+      await WakelockPlus.disable();
+    }
+  }
+
+  /// Re-applies the wakelock based on the current playback state. Called when
+  /// the app returns to the foreground, because Android drops the
+  /// keep-screen-on window flag across a background→foreground cycle and
+  /// nothing else restores it while playback continues.
```
PartyDonut: Function itself is clear enough let's remove comments where not really needed .

PartyDonut (dismissed):
Looks good to me and works fine. One small request to remove some comments.

lib/wrappers/media_control_wrapper.dart
aviadlevy: Done in a067f8c8

PartyDonut (approved):
Thanks for the quick changes 👍

PR #1063, opened 2026-07-20
feat: Previous and next chapter buttons

lexaproblem:
## Pull Request Description

Added previous and next chapter buttons in the video player. This has not been added to tv_player_controls.dart, but I will (probably) add this in another PR.

## Issue Being Fixed

None, besides my mild annoyance

## Screenshots / Recordings

image

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [x] Windows
- [ ] macOS
- [x] Web

## Checklist

### DIFF
--- a/lib/screens/video_player/video_player_controls.dart
+++ b/lib/screens/video_player/video_player_controls.dart
@@ -11,6 +11,7 @@ import 'package:iconsax_plus/iconsax_plus.dart';
 import 'package:screen_brightness/screen_brightness.dart';
 
 import 'package:fladder/models/item_base_model.dart';
+import 'package:fladder/models/items/chapters_model.dart';
 import 'package:fladder/models/items/media_segments_model.dart';
 import 'package:fladder/models/items/media_streams_model.dart';
 import 'package:fladder/models/media_playback_model.dart';
@@ -428,6 +429,7 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
                     ),
                   ),
                   previousButton,
+                  previousChapterButton(ref),
                   seekBackwardButton(ref),
                   IconButton.filledTonal(
                     iconSize: 38,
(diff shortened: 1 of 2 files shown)


lib/screens/video_player/video_player_controls.dart
```diff
@@ -685,6 +688,66 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
...
+        icon: const Icon(IconsaxPlusLinear.arrow_left_3),
+      );
+    });
+  }
+
+  void _previousChapter(WidgetRef ref) {
+    final chapters = ref.read(playBackModel.select((value) => value?.chapters)) ?? [];
+    if (chapters.isEmpty) return;
+    final position = ref.read(mediaPlaybackProvider).position;
+    final threshold = const Duration(milliseconds: 3000);
```
PartyDonut, line 721: There is a extension in `chapters_model.dart` lets move the next/previous lookups in there and only pass the required paramaters to fetch previous/next chapters.

That way we can re-use it when needed.

lib/screens/video_player/video_player_controls.dart
```diff
@@ -685,6 +688,66 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
...
+        break;
+      }
+    }
+    if (next != null) {
+      ref.read(videoPlayerProvider).seek(next.startPosition);
+    } else {
+      // fallback: load next video
+      final nextVideo = ref.read(playBackModel.select((value) => value?.nextVideo));
+      if (nextVideo != null) ref.read(playbackModelHelper).loadNewVideo(nextVideo);
+    }
```
PartyDonut, line 747: I'm not sure about loading the next video. Would it not be better to disable the next chapter button if there is no next chapter?

PartyDonut (changes requested):
Functionality looks good, have some small requests.

PR #1067, opened 2026-07-24
feat: hotkey for subtitle offset

lexaproblem:
## Pull Request Description

Adds hotkeys to offset subtitles forwards and backwards in time in increments of 250ms. The new hotkeys are left control + left/right arrow key. Adds an indicator to display how much the subtitles are offset in ms in the video player as well. The functionality appears to be somewhat buggy on the web client.

TODO?: Localization in video_player_settings.dart is currently just written in english, but I don't think it really matters that much as there's no way to display that text (as far as im aware)

## Issue Being Fixed

Resolves #670

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [x] Windows
- [ ] macOS
- [x] Web

## Checklist


lib/models/settings/video_player_settings.dart
```diff
@@ -63,6 +65,8 @@ enum VideoHotKeys {
       VideoHotKeys.takeScreenshot => context.localized.takeScreenshot,
       VideoHotKeys.takeScreenshotClean => context.localized.takeScreenshotClean,
       VideoHotKeys.toggleSubtitles => context.localized.toggleSubtitles,
+      VideoHotKeys.subtitleOffsetBackward => 'Subtitle backward',
```
PartyDonut, line 68: These should be localized in the app_en.arb base translations file.

lib/screens/video_player/components/video_player_subtitle_offset_indicator.dart
```diff
@@ -0,0 +1,79 @@
...
+        ),
+      ),
+    );
+  }
+}
+
+String _subtitleDelayLabel(Duration offset) {
+  final absMilliseconds = offset.inMilliseconds.abs();
+  final sign = offset.inMilliseconds >= 0 ? '+' : '-';
+  return 'Subtitle $sign${absMilliseconds}ms';
```
PartyDonut, line 78: Should also be translated

lib/screens/video_player/video_player_controls.dart
```diff
@@ -1000,6 +1002,8 @@ class _DesktopControlsState extends ConsumerState<DesktopControls> {
     final mediaSegments = ref.read(playBackModel.select((value) => value?.mediaSegments));
     final position = ref.read(mediaPlaybackProvider).position;
     final playing = ref.read(mediaPlaybackProvider.select((value) => value.playing));
+    final subtitlesEnabled = ref.read(playBackModel)?.mediaStreams?.defaultSubStreamIndex != null &&
```
PartyDonut, line 1005: We could just show the subtitle offset always, not sure we have to hide it if nothing is selected.

PartyDonut (changes requested):
Looks good to me thanks for helping out.

Left some small requests.

Currently the offset is remembered even when stopping/starting playback. This should probably be reset after the video is done playing?

lib/screens/video_player/video_player_controls.dart
alex-rans: I did it like this because thats the default behavior in VLC and Stremio

theodugautier:
@alex-rans  Thanks so much for the feature :) it's the only thing I'm really missing, and it makes the interface unusable for certain movies and series

PR #1075, opened 2026-08-01
fix: Persist subtitle font weight across app restarts

### LINKED ISSUE #1072: Subtitle customization setting resets on every startup
### Description

When I change the subtitle font from Normal to Bold, it works during that session, but as soon as I close the app and open again, the subtitle goes back to Normal font weight. Every time I open the app I have to go to subtitle style settings and set it back to Bold. Every other setting seems to be saving normally, just the font weight has this issue.

I tested both 0.10.3 (latest release) and 0.10.3 nightly from 2026-07-29. Also tested both Android TV 12, Android TV 14 and Android 14 (phone). The issue happens and is reproduceable on all mentioned scenarios.

Edit: Updated the Jellyfin Server to latest stable version (10.11.11) and the issue remains the same.

This is the most complete Jellyfin client I've found for my Android TV, thanks for making it happen. Besides some QoL stuff, this is the only thing really making it hard to totally replace the original client.

### Reproduction steps

```bash
1. Go to Settings > Player > Customize subtitles
2. Set font weight to Bold
3. Go back to the main screen and play any content with subtitles
4. Subtitle font is Bold as expected
5. Close the application and open again
6. Play any content with subtitles
7. Font weight won't be Bold and going to the customization settings again you can see it set back to Normal
```

### Screenshots

Image

### Logs

```bash

```

### Platform

Android, Android TV

### App Version

0.10.3

### Jellyfin server

10.11.6

Charlie284:
## Pull Request Description

Persist the selected subtitle font weight across app restarts.

`SubtitleSettingsModel` stores `FontWeight.value` (for example, `700` for Bold), but deserialization compared the stored number with the enum index. The restored value therefore fell back to Normal. Deserialization now compares against `FontWeight.value`, with regression coverage for a Bold serialization round trip.

## Issue Being Fixed

Fixes #1072

## Screenshots / Recordings

Not applicable. This change only affects settings persistence.

## Tested On

- [ ] Android
- [x] Android TV — Android 14/API 34 ARM64 emulator
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist

- [x] No new package was added.
- [x] All changes are related to the issue at hand.

## Testing

- `flutter analyze`
- `flutter test test/pip_manager_test.dart test/subtitle_settings_model_test.dart`
- `flutter build apk --debug --split-per-abi --flavor production`
- Installed the ARM64 production debug APK on an Android TV 14 emulator
- Verified `fontWeight:700` remained persisted through two cold launches and force-stops


PartyDonut (approved):
Thanks for fixing this.

Removed the test as currently we don't really need/run any tests. Which I probably should add at some point 😓 .

Issue #912, opened 2026-04-06
🐛 Ass subtitles won't show when using transcoded playback


DestinyofYeet:
### Description

Videos which are being viewed transcoded instead of direct won't show subtitles.

### Reproduction steps

```bash
1. Play a video (direct playback)
2. Subtitles are there
2. Select some bitrate
3. No more subtitles (transcoded playback)
```

### Logs

```bash

```

### Platform

Linux

### App Version

0.10.2

### Jellyfin server

10.11.6

rlad78:
Just wanted to note I'm seeing the same issue on 0.10.3 for Linux (Arch, using bin from releases), Android (16, mobile), and macOS (26.3.1). Jellyfin server is on version 10.11.8. Can't yet validate that my issue is directly related to transcoding, as all of my streams have to be transcoded due to bandwidth restrictions. I'll check later when I get home if I get different results.

I'm not too well versed in ffmpeg, but a cursory bit of digging leads me to believe that the subtitles are being excluded from the stream. Forgive me if I'm being a bit naive, but see the `-map -0:s` flag below:

(from Jellyfin transcoding log)
[...]

Issue #861, opened 2026-03-19
🐛 Subtitle size does not scale with video resolution (TV android app)


bientavu:
### Description

The new subtitle configuration implemented in the android TV app in [v0.10.1](https://github.com/DonutWare/Fladder/releases/tag/v0.10.1) is great, but there appears to be a scaling issue depending on the source media.

Expected Behavior: Subtitles should maintain a consistent relative size across different video resolutions.

Actual Behavior: The absolute size of the subtitles seems to change based on the video resolution. For instance, the same configuration results in much smaller subtitles on a 4K video compared to a 1080p video.

I believe it might not only be for subtitle size, but also height offset.

### Reproduction steps

```bash
1. Go to settings and configure a specific subtitle size
2. Play a video with a 1080p resolution and see how it's setup
3. Play a 4K video, and see that the subtitles render smaller
```

### Logs

```bash

```

### Platform

Android TV

### App Version

v0.10.2

### Jellyfin server

10.11.6

ditmarvisser:
I can also add that the subtitle height is dependent on the lower bound of the video. So a 16:9 video has absolute lower subtitles than a 21:9 video. But they both have the same height relative to the bottom of the video.

Issue #473, opened 2025-08-24
🐛 Subtitle Not Loading When External Audio Transcoding


ajanjairam:
### Description

Subtitle is not displaying even if it is selected and this occurs only in transcoding with external audio reason.

URL: <host:port>/videos/<video_id>/master.m3u8?DeviceId=<device_id>&MediaSourceId=<media_source_id>&VideoCodec=h264&AudioCodec=aac&AudioStreamIndex=0&VideoBitrate=119933383&AudioBitrate=66617&AudioSampleRate=48000&MaxFramerate=23.976025&PlaySessionId=<session_id>&api_key=<api_key>&TranscodingMaxAudioChannels=2&RequireAvc=false&EnableAudioVbrEncoding=true&Tag=<tag_id>&SegmentContainer=ts&BreakOnNonKeyFrames=False&h264-level=40&h264-videobitdepth=8&h264-profile=high&h264-audiochannels=2&aac-profile=lc&TranscodeReasons=AudioIsExternal

### Reproduction steps

```bash
1. Go to a Video with external audio
2. Click on Play
3. Subtitle is selected but not displaying
```

### Logs

```bash

```

### Platform

Android

### App Version

0.7.5

wtfolivr:
+1 on this issue for 0.8.0 Android based device.

MPV Shim and JF Browser Player transcode and subtitle works just fine but never Fladder v0.8.0

PR #1077, opened 2026-08-06
fix(Android): Render ASS subtitles on the native player

### LINKED ISSUE #919: 🐛 External ASS subtitles aren't playing on Android TV
### Description

I have a series with .ass subtitles for each episode, named accordingly. They are played with no issue in jellyfin web or in any fladder client i tested (android, desktop) and EVEN on this exact tv when in tablet layout (it was detected this way after i opened fladder right after the tv reboot, another bug?). However, when the app is opened on tv with correct television layout, i can choose the subs but there will be nothing on screen while playing media

Archive with subtitles file i'm having issue with: [subs.zip](https://github.com/user-attachments/files/26519461/subs.zip)

### Reproduction steps

```bash
1. Have correctly named .ass subs along with episode
2. Rescan library
3. Make sure these subs are working in jellyfin webui
4. Play the same media with the same subs in fladder on android tv and see no subtitles
```

### Screenshots

_No response_

### Logs

```bash
Nothing in logs
```

### Platform

Android TV

### App Version

v0.10.2

### Jellyfin server

10.11.8, linuxserver's image

bilbofroggins:
## Problem

ASS/SSA subtitles never render on the Android TV / leanback player. The track appears in the picker, can be selected, and simply draws nothing — no error, no crash, nothing in logs. The same file plays with subtitles correctly on desktop, on web, and even in the same app on the same device in tablet layout.

Fixes #919.

## Why only Android TV

`PlayerOptions.available` (`lib/models/settings/video_player_settings.dart:177`) returns only `nativePlayer` in leanback mode, so TV playback runs through `VideoPlayerActivity` → `android/.../player/ExoPlayer.kt` and renders ASS with **libass** via `io.github.peerless2012:ass-media`. Every other platform uses mpv/mdk. Three different subtitle renderers, and only the libass-on-Android one was broken.

## Root cause

libass is built for Android without a font provider, and `ass-kt` initialises the renderer with:

```c
// lib_ass_kt/src/main/cpp/AssKt.c
ass_set_fonts(assRenderer, NULL, "sans-serif", ASS_FONTPROVIDER_FONTCONFIG, NULL, 1);
```

It requests the **fontconfig** provider, which does not exist in that build, and passes **no default font path**. libass therefore starts with zero fonts:

```
W SubtitleRenderer: can't find selected font provider
W SubtitleRenderer: fontselect: failed to find any fallback with glyph 0x0 for font: (Arial, 700, 0)
```

The failure is silent because of where it lands in the library. `AssSubtitleParser.parse()` rasterises each event and only emits a cue from *inside* `frames?.images?.let { … }`. With no font, `renderFrame()` returns a frame containing no images, so that block never runs and **not one cue is ever handed to ExoPlayer**.

SRT is unaffected: media3 renders it with Android's own `Typeface` stack and never touches libass. That asymmetry is what makes the bug look mysterious from the outside.

## Why registering fonts isn't enough on its own

libass resolves a font in a fixed order (`ass_font_select`, `ass_fontselect.c`):

| # | Step | Result here |
| --- | --- | --- |
| 1 | family the script asks for | `Arial`, `Nirmala UI`, … not on the device |
| 2 | `family_default` | `"sans-serif"`, per the call above |
| 3 | provider's `get_fallback` | no provider exists |
| 4 | `path_default` | `NULL`, and `ass-kt` cannot set it |

Steps 1, 3 and 4 are unreachable from app code, so a font registered as `Droid Sans Fallback` is loaded and then never looked at. Step 2 is the only lever — which means the fallback font has to literally *be* named `sans-serif`.

## The fix

`AssFonts.install()` registers fonts through `Ass.addFont()` and rewrites the fallback font's sfnt `name` table so its family reads `sans-serif`. `"sans-serif"` is shorter than `"Droid Sans Fallback"`, so the records are patched in place — no table resizing, no offset changes, file size unchanged.

The fallback is `assets/mp-font.ttf` (Droid Sans Fallback), already shipped for mpv's `subtitleFontFile` (`lib/wrappers/players/base_player.dart:12`), so CJK is covered and the native player now falls back to the same font as every other platform. `Roboto` and `DroidSans` are also registered under their real names so scripts naming them resolve at step 1.

Two supporting changes:

- **`buildWithAssSupport()` is inlined** in `ExoPlayer.kt`. It constructs the `AssHandler` internally and never returns it, and that handler owns the `Ass` instance fonts attach to. Same wiring, same render type — and as a side benefit the helper no longer silently replaces the configured `dataSourceFactory` with a plain `DefaultDataSource.Factory`.
- **`ass-kt` is declared explicitly** in `build.gradle`, because `ass-media` only depends on it at *runtime* scope, leaving `Ass` off the compile classpath.

## Testing

Verified on a Sony BRAVIA 4K VH2 (Android 12, `leanback_only`, armeabi-v7a) with a direct-played mkv carrying an embedded ASS track and **no font attachments** — the case that was previously guaranteed to fail:

```
fontselect: Using default font family: (Arial, 700, 0) -> DroidSansFallback
fontselect: (Nirmala UI, 700, 0) -> DroidSansFallback
```

Subtitles render with their ASS styling intact — the source's `{\c&H00D8FF&\3c&H054D8B&}` shows up as gold text with a dark blue outline, so colours and outlines survive rather than degrading to plain text. An SRT control cut of the same file was checked before and after to confirm no regression on the non-libass path.

## Related, but deliberately not claimed

- **#1009** overlaps — it also adds font fallback handling, bundled with transcoding track-selection changes, and **will conflict with this in `ExoPlayer.kt`**. This PR is deliberately narrower: one root cause, independently verified on hardware. Happy to rebase around whichever lands first.
- **#861** (subtitle size doesn't scale with video resolution on TV) is adjacent and *not* fixed here. `AssRenderType.LEGACY` sizes libass's frame to the video dimensions rather than the display; moving to the overlay renderer would address it.
- **#912** is Linux/mpv, unrelated to this code path.
- **#473** is transcoding: Jellyfin delivers the subtitle as a sidecar URL, and `VideoPlayerImplementation.open()` only side-loads tracks with `external=true`, which is false for a stream embedded in the source. It is dropped before libass is involved. This fix is a *prerequisite* for that working, not a fix for it.

## Upstream notes

- The real defect is in `ass-kt`: requesting a provider that does not exist in the Android build while passing no default font makes libass unusable unless the host app happens to register a font named `sans-serif`. Worth filing upstream; this workaround can be dropped if `ass_set_fonts` ever becomes configurable.
- `ass-media` is pinned to **0.3.0**. Font *attachments* embedded in an mkv also appear not to load on that version (a test file with a muxed font rendered nothing); peerless2012/libass-android#63 "Fix font loading when attachments precede tracks in MKV" shipped in **0.4.0**. A version bump is worth a separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -139,6 +139,9 @@ dependencies {
     implementation("androidx.media3:media3-exoplayer-hls:$media3_version")
     implementation("org.jellyfin.media3:media3-ffmpeg-decoder:$media3_version")
     implementation("io.github.peerless2012:ass-media:0.3.0")
+    // ass-media only depends on ass-kt at runtime scope, so Ass/AssRender are not on
+    // the compile classpath. We need Ass.addFont() to give libass a font (see AssFonts).
+    implementation("io.github.peerless2012:ass-kt:0.3.0")
 
     //UI
     implementation("io.github.rabehx:iconsax-compose:0.0.5")
--- a/android/app/src/main/kotlin/nl/jknaapen/fladder/player/ExoPlayer.kt
+++ b/android/app/src/main/kotlin/nl/jknaapen/fladder/player/ExoPlayer.kt
@@ -42,7 +42,10 @@ import androidx.media3.extractor.ts.TsExtractor
 import androidx.media3.ui.AspectRatioFrameLayout
 import androidx.media3.ui.CaptionStyleCompat
 import androidx.media3.ui.PlayerView
-import io.github.peerless2012.ass.media.kt.buildWithAssSupport
+import io.github.peerless2012.ass.media.AssHandler
+import io.github.peerless2012.ass.media.kt.withAssMkvSupport
+import io.github.peerless2012.ass.media.kt.withAssSupport
+import io.github.peerless2012.ass.media.parser.AssSubtitleParserFactory
 import io.github.peerless2012.ass.media.type.AssRenderType
 import kotlinx.coroutines.delay
 
@@ -52,6 +55,7 @@ import nl.jknaapen.fladder.messengers.properlySetSubAndAudioTracks
 import nl.jknaapen.fladder.objects.PlayerSettingsObject
 import nl.jknaapen.fladder.objects.VideoPlayerObject
 import nl.jknaapen.fladder.utility.AllowedOrientations
+import nl.jknaapen.fladder.utility.AssFonts
 import nl.jknaapen.fladder.utility.conditional
 import nl.jknaapen.fladder.utility.getAudioTracks
 import nl.jknaapen.fladder.utility.getSubtitleTracks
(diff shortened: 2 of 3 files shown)


PartyDonut:
This seems like a hacky-fix to what I'm not even sure is the exact problem.

Ass fonts seem to render just fine on my Chromecast 4k. Is this specifically fixing external ass subtitles, or just fixing devices that are missing the required "default" fonts?

bilbofroggins:
> This seems like a hacky-fix to what I'm not even sure is the exact problem.
> 
> Ass fonts seem to render just fine on my Chromecast 4k. Is this specifically fixing external ass subtitles, or just fixing devices that are missing the required "default" fonts?

The problem, at least for me, is not external ass subtitles - it's embedded ones. And I'm not sure about the device missing fonts but I think it's about how they are loaded. I'd be curious to know what type of media you are trying to play. Sometimes the fonts are included into the subtitles, which might be the difference we're seeing. I was testing with this video, which has english subs built-in:
`yt-dlp 'https://archive.org/details/os-038-k-38-cyber-soldier-polygon/OS001_K01+-+Pok%C3%A9mon!+I+Choose+You!.mkv'`

PR #1083, opened 2026-08-10
fix(Android): support local network access on Android 16+

aviadlevy:
## Pull Request Description

Adds support for Android 16's Local Network Protection, which requires apps to explicitly declare and request the `ACCESS_LOCAL_NETWORK` permission before they can talk to devices on the local network. Fladder never declared it, so on Android 16+ the system quietly blocked every connection to a Jellyfin server on the user's own network.

### Changes:

- **Declares `android.permission.ACCESS_LOCAL_NETWORK`** in the manifest. Older Android versions don't know the permission and ignore it, so nothing changes  there.
- **Requests it where it's needed** — when a local URL is saved in settings, and  on startup when one is already configured (existing users would otherwise never  be prompted, since they never re-save their URL).
- **Handles denial.** Android only prompts once, so after a denial there is no way  back from inside the app. The profile settings page now shows a warning that  opens Fladder's system settings page and re-checks the permission on resume —  the same pattern the page already uses for battery optimization.
- **Makes the local server probe more robust.** It's re-run once access is  granted, rather than latching on the result of a probe that ran while traffic  was still blocked, and its timeout goes from 1s to 3s, which a cold Wi-Fi  handshake can exceed.

The platform side is a small Pigeon host API (`pigeons/local_network_access.dart`), matching the existing `battery_optimization` setup.

### AI assistance

Root-cause investigation and the patch were done with Claude Code; diagnosis, fix design, and on-device validation were human-reviewed and -owned.

## Issue Being Fixed

On Android 16+ devices, Fladder cannot reach a Jellyfin server on the local network. The "Local Jellyfin URL" indicator stays red and the app reports **Offline**, even though the server is reachable from every other device and browser on the same LAN.

## Tested On

- [x] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Web

## Checklist


schembriaiden:
Same as this: https://github.com/DonutWare/Fladder/pull/1112, I think

PartyDonut:
Yes, completely forgot this one sorry about that @aviadlevy. 

#1112 does also fix a few related things. So closing this one for now.

Issue #786, opened 2026-02-23
🐛 Fladder shows "offline" banner even when server is online


viggy96:
### Description

Fladder oftentimes shows the "offline" batter along the top of the application even when the server is online and connected. I can be actively browsing through my library and it will still show offline. On Linux I have to close and reopen the app to get rid of the banner. On Android, I have to manually refresh the page to get it to go away. 

### Reproduction steps

```bash
1. Open app
2. Browse for a while
3. See offline banner
4. See error
```

### Logs

```bash

```

### Platform

Linux

### App Version

0.10.1

### Jellyfin server

10.11.6

Issue #623, opened 2025-11-17
🐛 JF 10.11.3 Update - Fladder Shows As Offline


SABRE1982:
### Description

With the release of the 10.11.3 update, Fladder is now unable to access the server. It shows as being offline. This has been tested on macOS and Android. 

### Reproduction steps

```bash
1. Go to 'Your Server'
2. Sign in
3. See error
```

### Screenshots

<img width="1470" height="375" alt="Image" src="https://github.com/user-attachments/assets/d8a0092f-4eb1-4831-a6b9-cfa021c425ed" />

### Logs

```bash

```

### Platform

macOS

### App Version

v0.8.0

### Jellyfin server

10.11.3

SABRE1982:
I should have noted that I am able to login using other clients (official, Streamyfin) without issues. It appears to be isolated to Fladder only. 

I have also tried signing in from scratch.

SABRE1982:
Logs:

[SEVERE  |  Flutter error: Exception: Failed to make request
ClientException: Connection closed while receiving data, uri=REDACTED/Users/ca7bc3d15e3747baa69425a97f12ae03/Items/Latest?parentId=f137a2dd21bbc1b99aa5c0f6bf02a805&fields=ParentId&fields=MediaStreams&fields=MediaSources&fields=CanDelete&fields=CanDownload&fields=PrimaryImageAspectRatio&fields=Overview&imageTypeLimit=1&enableImageTypes=Primary&enableImageTypes=Backdrop&enableImageTypes=Thumb&limit=16, 2025-11-22T22:35:25.299608

#0      JellyRequest.intercept (package:fladder/providers/api_provider.dart:81)
<asynchronous suspension>
#1      InterceptorChain.proceed (package:chopper/src/chain/interceptor_chain.dart:46)
<asynchronous suspension>
#2      RequestConverterInterceptor.intercept (package:chopper/src/interceptors/request_converter_interceptor.dart:26)
<asynchronous suspension>
#3      InterceptorChain.proceed (package:chopper/src/chain/interceptor_chain.dart:46)
<asynchronous suspension>
#4      Call.execute (package:chopper/src/chain/call.dart:56)
<asynchronous suspension>
#5      ChopperClient.send (package:chopper/src/base.dart:175)
<asynchronous suspension>
#6      ViewsNotifier.fetchViews.<anonymous closure> (package:fladder/providers/views_provider.dart:50)
<asynchronous suspension>
[...]

PR #1114, opened 2026-09-04
fix: Never commit offline from the OS connectivity signal alone

matijarma:
## Problem

On Windows desktop the app gets stuck in offline mode while the server is perfectly reachable. The banner appears at launch or while browsing and does not go away until the app is restarted. Because `offlineStateProvider` gates the service layer, the library looks empty, and playing from the home screen fails with "no compatible media found": `playback_model.dart` goes straight to local synced files when offline and never contacts the server.

Related: #786 (closed, same symptom), #623.

## Root cause

The platform connectivity signal is treated as the verdict. `_handleHardwareChange` commits `ConnectionState.offline` the moment the OS reports `none`, and an unclassifiable transport (`other`, `bluetooth`, an empty list) is mapped to offline as well.

On Windows that signal is wrong on healthy networks. The plugin's native code only counts adapters that the OS Network List Manager reports as `IsConnectedToInternet`, a verdict derived from Microsoft's NCSI probe and transiently false right after launch, during DHCP renewal, or when Hyper-V/WSL/VPN virtual adapters churn. Measured on my machine: the exact endpoint the probe uses answered 5/5 times in ~0.15 s while the app sat in offline mode.

## Relationship to f351ae62

This is rebased on f351ae62 ("Another connectivity check pass") and keeps all of it: re-checking after every request, `waitForProbe()`, the local-network permission gate. That pass makes recovery faster once the OS signal is right again, but while the signal is wrong it re-commits offline after every response, so the false-negative case is unchanged. This PR only changes what happens when the signal says offline or unknown.

## Change

- The result mapping is a pure function (`mapConnectivityResults`) that reports `other`/`bluetooth`/empty as *unknown* instead of offline. Unit-tested.
- Offline is never committed from the hardware signal alone. A usable transport is recorded (it drives bitrate selection); offline or unknown is verified by the existing probe, using the last known transport as the candidate state.
- While offline, the probe retries with backoff from 15 s to a 2-minute cap, so a false negative recovers on its own. Desktop platforms rarely emit the lifecycle events that would otherwise trigger a re-check.
- Probe timeout 2 s → 5 s. It now guards the remote URL as well, a cold server or slow link routinely exceeds two seconds, and a timeout here is what commits offline.

Trade-off: a genuine loss of connectivity registers once the probe gives up (500 ms debounce plus up to two 5 s attempts) rather than instantly.

## Testing

- Windows 10 (22H2) desktop, release build, Jellyfin 10.10.7 over HTTPS on a non-standard port, no local URL configured. On 0.11.0 the offline banner appeared at almost every launch and stuck; with the "OS signal is a hint" change it no longer appears, and playing from the home screen works.
- Unit tests for `mapConnectivityResults` cover vpn/ethernet/wifi/mobile precedence, explicit `none`, and the regression cases (`other`, `bluetooth`, empty list → unknown, not offline).
- `flutter analyze --fatal-infos` clean; `flutter test` green except `test/widget_test.dart`, the default template counter test, which fails on `develop` too.

Reproducing the original bug deterministically: add `127.0.0.1 www.msftconnecttest.com` to the hosts file. Windows then reports "no internet" while the server stays reachable, which is exactly the condition that pins the app offline.


matijarma:
Rebased onto `develop` after f351ae62 ("Another connectivity check pass") landed.

The two changes rewrote the same file, and the "Update branch" merge GitHub produced was syntactically broken (the CI failure above was against that merge commit, not the PR itself). This is now a single commit on top of f351ae62 that keeps all of it and only changes what happens when the OS reports offline or an unclassifiable transport: verify with the existing probe instead of committing, and retry with backoff while offline. Details in the description.

matijarma:
just wanted to report after 10 days of usage - experience is wildly better, perfect actually.

PR #1120, opened 2026-09-06
feat: Custom headers for servers behind an authenticating proxy

Wasabules:
## Pull Request Description

Adds per-server custom HTTP headers to the Jellyfin connection, mirroring what Seerr already supports in Fladder. The use case is a Jellyfin instance behind Cloudflare Access / Zero Trust, which rejects every request that does not carry `CF-Access-Client-Id` and `CF-Access-Client-Secret`, but any header authenticated reverse proxy works the same way.

**Where they are set**

- Login screen → the **Advanced** button. It moved next to the server field: a server behind Cloudflare Access cannot be reached at all until its headers are set, and the button used to only appear *after* a successful connection. The dialog keeps the Seerr URL field it already had.
- Settings → Profile → Advanced → **Custom headers**, for an account that is already logged in.

The editor offers a **Cloudflare Access** preset with a Client ID and a Client secret field carrying the right header names, so nobody has to type `CF-Access-Client-Id` by hand, and a **Custom** preset for free-form entry. A server that already has Cloudflare headers configured opens on that preset.

The headers live on `CredentialsModel`, so they are per account and are persisted next to the token.

**Where they are applied**

- Every Jellyfin API call, through `CredentialsModel.header`
- The server probe done before login, and the local-URL reachability probe
- Cached images and trickplay tiles
- Video and music playback (libMPV and libMDK)
- Downloads, and everything offline sync fetches (artwork, chapter images, external subtitles)
- The background update-notification worker

**Not leaking them**

`ServerCustomHeaders` only hands the headers out for URLs whose host matches the server URL or the local URL. Seerr/TMDB artwork travels through the same image cache, and those requests must never carry the user's Cloudflare credentials. `test/custom_headers_test.dart` covers that.

**Two things that came out of testing this against a real Cloudflare Access instance**

- A server behind Cloudflare Access answers the login probe with a 302 to its identity provider, which surfaced as *"Invalid URL"* and sent me looking at a perfectly good address. That case is now recognised from the `Www-Authenticate` challenge or the redirect target, and says the server wants credentials of its own instead.
- The buttons flanking the server field are `AspectRatio(1)` children of an `IntrinsicHeight` row, so their width follows the row height. A multi-line error rendered inside the field grew the row, grew the buttons, and left the field no width at all — it disappeared and the row overflowed. The error now sits in a message box under the row and the buttons carry a width cap, so a longer translation cannot reproduce it.

**Known gap**

The Android native player backend (used in lean-back mode) is not covered. It opens its URL on the Kotlin side through a Pigeon API that has no headers field, so wiring it up needs a native change. Happy to fold that in here if you'd rather not have the gap.

## Issue Being Fixed

Resolves the request in discussion #823 — Jellyfin instances exposed through a Cloudflare Tunnel and protected by Cloudflare Access cannot be reached at all today, because Cloudflare intercepts every request, `/System/Info/Public` included, before it reaches Jellyfin.

## Screenshots / Recordings

_To follow._

## Tested On

- [x] Android
- [ ] Android TV
- [ ] iOS
- [ ] Linux
- [x] Windows
- [ ] macOS
- [ ] Web

Tested end to end on Android against a Jellyfin server behind Cloudflare Access with a service token: reaching the server, logging in, artwork loading and video playback all work through the proxy. Also `flutter analyze` clean and `dart format --line-length=120` on every touched file.

## Checklist

- [x] If a new package was added, did you ensure it works for all supported platforms? Is the package well maintained — no new packages; this only uses `http`, `flutter_cache_manager`, `media_kit` and `video_player`, which are already dependencies.


Wasabules:
https://www.youtube.com/shorts/XTJB-h46cUs

Tested on Android: authentication and video playback work with Cloudflare Zero Trust using an access token. Please note that the tokens shown in the video have been revoked. 
I'm going to test the addition on Windows and Linux.

Wasabules:
Tested on Windows as well; it works.


https://github.com/user-attachments/assets/64b91631-88e4-4286-aa8c-76837ff86bb4


However, I discovered a bug related to loading playlists; since my Jellyfin setup is quite large, it overloads my Jellyfin instance. I have opened an  issue #1121 .

Please note that the tokens shown in the video have been revoked.

PR #1132, opened 2026-09-10
chore: remove scheduled build

schembriaiden:
## Pull Request Description

Removes the scheduled build since now the build workflow runs on every commit to develop, meaning one less build.


PartyDonut (approved):
Yup good idea.

PR #1139, opened 2026-09-16
fix: duplicate app name in title bar

schembriaiden:
## Pull Request Description

In https://github.com/DonutWare/Fladder/pull/1005 `window_title_provider.dart` had some code changed which resulted in the double Fladder in the title bar when watching a series/movie. This code change aims to return it back to its original form which stopped this from happening. I am not entirely sure if this was intended as I do not have music on my jellyfin server to verify, so maybe you could check the music yourself and see if that is affected, because with these changes the word `Tjilp` will probably no longer be visible when listening to music, and if that is wanted we just need to make a distinction between the normal version and the music version of the app, so let me know.

## Issue Being Fixed

Fixes the double Fladder in the title bar

## Screenshots / Recordings

Before this PR:

image

After this PR:

image

## Tested On

- [ ] Android
- [ ] Android TV
- [ ] iOS
- [x] Linux
- [ ] Windows
- [x] macOS
- [ ] Web

## Checklist


PartyDonut:
Did not account for how linux (or windows for that matter) shows the title.

The reason for this change was how Macos shows multiple windows under the same process.

Current build
image

So we should probably keep it as is for MacOS. And have a look whether windows requires the title or not.

It would be preferred to change the title of the window to `Tjilp` when playing music on all platforms if possible.

schembriaiden:
I can test it on windows, mac and linux, so I will check it out later no problem :)

schembriaiden:
@PartyDonut I have tested it on both Mac and Linux (both music dashboard and normal dashboard), but I haven't been able to test it on Windows because it requires Visual Studio 2019 due to the flutter sdk being a bit old and I have the 2026 version installed. So maybe you could test that if you are able too.

I made them all look like the screenshot you sent me for MacOS.

I also changed the dot with dashes for the movie and season player so that it matches the music one.

I also made a slight change to the podfile so that we only edit one place with the `macos_deployment_target`.

